PR #20, opened 2013-09-10
JSON Reporter

tsenart:
As a way of integrating Vegeta CLI with CI environments such as Jenkins, a more machine readable version of the text reporter would be of good use.

### DIFF
--- a/lib/metrics_test.go
+++ b/lib/metrics_test.go
@@ -43,7 +43,7 @@ func TestNewMetrics(t *testing.T) {
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


andrei-m:
@tsenart I'm looking into how to best integrate the JSON report format into Jenkins. Did you use a Jenkins plugin with this format to display test results?

tsenart:
Hi @andrei-m,

I never integrated Vegeta  with a Jenkins based CI pipeline. It shouldn't be too hard though!

PR #91, opened 2014-11-06
Add -keepalive flag to toggle persistent conns

### LINKED ISSUE #89: Add disable-keepalive flag
The service I'm testing almost exclusively deals with short-lived, single request connections. With no option to disable keepalive I cannot run tests which properly simulate my production environment.

Would you accept a PR with a `-disable-keepalive` flag?

jpfuentes2:
The flag is enabled by default.

closes #89


yhafri:
@tsenart can you please explain how to disable _keepalive_?
Should i pass this option to vegeta like this:
```
echo | ... vegeta attack -keepalive=0 ...
```
I want vegeta to send 1 request to my server and close the connection immediately.
No more than 1 request per connection at a time.

Thank you.

tsenart:
Yesm that's correct, you should pass `-keepalive=false` to disable keep alive, so every new request will establish a new TCP connection.

PR #96, opened 2014-11-12
Added Proxy as an option

MrSander:
We use this tool on a server remotely and also we need to use proxy but proxy wasn't set up on the server. In this case we can provide proxy settings via command line as an argument.


hasantayyar:
I couldn't make `HTTPS_PROXY` work.
Is the ssl proxy flag `HTTP_PROXY ` supported?

PR #114, opened 2015-03-11
Always read response body for each hit

### LINKED ISSUE #113: If chunked encoding is used, vegeta reports 0 bytes transferred
I just instrumented Vegeta [here](https://github.com/tsenart/vegeta/blob/670b8caab6c833181cae772c293664017d035bf1/lib/attack.go#L207) and discovered that my ContentLength is set to -1.

Vegeta is reporting 0 BytesOut, and I would really like to know the size. Could vegeta measure it by doing an `io.Copy(ioutil.Discard, req.Body)`, or would that have a negative performance impact?

tsenart:
With chunked encoding, an HTTP server sends the response headers right
away and then streams the body chunks which can be lazily generated.
Without chunked encoding, network transfer time of the body wasn't
previously being accounted for.

Fixes #113

### DIFF
--- a/lib/attack.go
+++ b/lib/attack.go
@@ -3,6 +3,8 @@ package vegeta
 import (
 	"crypto/tls"
 	"fmt"
+	"io"
+	"io/ioutil"
 	"net"
 	"net/http"
 	"strings"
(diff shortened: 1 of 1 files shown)


pwaller:
LGTM. Do we also want to count the headers in the transfer count?

On 11 March 2015 12:16:32 GMT+00:00, "Tomás Senart" <notifications@github.com> wrote:
>/cc @pwaller
>
>---
>Reply to this email directly or view it on GitHub:
>https://github.com/tsenart/vegeta/pull/114#issuecomment-78252020

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

tsenart:
That'd change the semantics of `Bytes{In,Out}`, which is the number of bytes in response and request body, so not for these fields in particular.

PR #116, opened 2015-03-17
Swap "of need" to "of a need" in README

ciarand:
sorry

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # Vegeta [![Build Status](https://secure.travis-ci.org/tsenart/vegeta.png)](http://travis-ci.org/tsenart/vegeta)
 
-Vegeta is a versatile HTTP load testing tool built out of need to drill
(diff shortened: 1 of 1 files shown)


tsenart:
What are you sorry for? :-)

ciarand:
stupid PRs

tsenart:
I wouldn't call descriptive PRs stupid. Although I'm fluent in English, it's not my native language. Would you please explain me why the form you suggest is more correct than the one before?

ciarand:
Sure! Here's the original for reference (emphasis mine):

>Vegeta is a versatile HTTP load testing tool built **out of need** to drill HTTP services with a constant request rate

The problem is that the phrase "need to X" is a noun phrase, so you need a noun there. In the case above the `a` becomes the target of the noun phrase so the sentence becomes ok.

tsenart:
Cool, thanks for elucidating :-)

ciarand:
You're welcome, thanks for the awesome tool!

PR #125, opened 2015-05-18
Allow case-sensitive headers.

### LINKED ISSUE #124: Case Sensitive headers
Hi there.
A quick issue that I've encountered today using vegeta binary (v5.6.3 amd64). Considering the following target file:

image

If you proxy the requests sent by vegeta, you got something like this (Headers are sorted and Capitalized) :
image
 
I know that RFC 2616 specifies that headers should be case-insensitive, but the REST application that I'm testing is sensitive.
I understand that http.header's function add and set canonicalize headers, but is there any quick hack to make vegeta respect cases?
Thanks !

GeertJohan:
Fixes #124

### DIFF
--- a/attack.go
+++ b/attack.go
@@ -186,7 +186,7 @@ func (h headers) Set(value string) error {
 	if key == "" || val == "" {
 		return fmt.Errorf("header '%s' has a wrong format", value)
 	}
-	h.Add(key, val)
+	h.Header[key] = append(h.Header[key], val)
 	return nil
 }
 
--- a/attack.go
+++ b/attack.go
@@ -177,6 +177,7 @@ func (h headers) String() string {
 	return buf.String()
 }
 
+// Set implements the flag.Value interface for a map of HTTP Headers.
 func (h headers) Set(value string) error {
 	parts := strings.SplitN(value, ":", 2)
 	if len(parts) != 2 {
@@ -186,6 +187,9 @@ func (h headers) Set(value string) error {
 	if key == "" || val == "" {
 		return fmt.Errorf("header '%s' has a wrong format", value)
 	}
+	// Add key/value directly to the http.Header (map[string][]string).
+	// http.Header.Add() cannonicalizes keys but vegeta is used
+	// to test systems that require case-sensitive headers.
 	h.Header[key] = append(h.Header[key], val)
 	return nil
 }
--- a/attack_test.go
+++ b/attack_test.go
@@ -1,0 +1,28 @@
+package main
+
+import (
+	"net/http"
+	"reflect"
+	"testing"
+)
+
+func TestHeadersSet(t *testing.T) {
+	h := headers{
+		Header: make(http.Header),
+	}
+	for i, tt := range []struct {
+		key, val string
+		want     []string
+	}{
+		{"key", "value", []string{"value"}},
+		{"key", "value", []string{"value", "value"}},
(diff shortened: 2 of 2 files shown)


attack.go
```diff
@@ -186,7 +186,7 @@ func (h headers) Set(value string) error {
 	if key == "" || val == "" {
 		return fmt.Errorf("header '%s' has a wrong format", value)
 	}
-	h.Add(key, val)
+	h.Header[key] = []string{val}
```
tsenart: This is equivalent to `.Set` not `.Add`. Should be: `h.Header[key] = append(h.Header[key], val)`

GeertJohan:
I thought this was a bug since your method is called `Set`, not `Add`.
vegeta's header.Set also accepts a full header key:value pair, so it makes sense to overwrite the complete combination. Is it currently being used to add multiple values to the same key?

attack.go
GeertJohan: I thought this was a bug since your method is called `Set`, not `Add`.
vegeta's `headers.Set` also accepts a full header `key:value` pair, so it makes sense to overwrite the complete combination. Is it currently being used to add multiple values to the same key?

attack_test.go
```diff
@@ -0,0 +1,29 @@
...
+		Header: make(http.Header),
+	}
+	for k, v := range testvalues {
+		err := h.Set(k + ":" + v)
+		if err != nil {
+			t.Errorf("error setting header %s:%s: %s", k, v, err)
+		}
+	}
+
+	for k, v := range testvalues {
```
tsenart: Could we make this one loop only? Something like this:
```go
for i, tt := range []struct{
  key, val string
  want []string
} {
  {"key", "value", {"value"}},
  {"key", "value", {"value", "value"}},
  {"Key", "Value", {"Value"}},
  {"KEY", "VALUE", {"VALUE"}},
} {
  if err := h.Set(tt.key + ": " + tt.val); err != nil {
    t.Error(err)
  } else if got := h.Header[tt.key]; !reflect.DeepEqual(got, tt.want) {
    t.Errorf("test #%d, '%s: %s': got: %+v, want: %+v", i, tt.key, tt.val, got, tt.want)
  }
}
```

attack.go
tsenart: The method is called `Set` in order to implement the `flag.Value` interface: http://golang.org/pkg/flag/#Value
Yes, it's used to set an arbitrary number of headers that have the same key which are valid HTTP semantics.

GeertJohan: Ah, I get it now. Cool I'll update the PR.

attack_test.go
GeertJohan: Sure!

GeertJohan:
Just a second, I'll also add a description on headers.Set for future readers.

GeertJohan:
Done

attack.go
```diff
@@ -177,6 +177,9 @@ func (h headers) String() string {
 	return buf.String()
 }
 
+// Set implements the flag.Value interface. It's used to set an
+// arbitrary number of headers that have the same key which are
+// valid HTTP semantics.
```
tsenart: Sorry, I'd rather not have this comment here. Maybe: `Set implements the flag.Value interface for a map of HTTP Headers.`

GeertJohan: Done

tsenart:
Awesome, thanks!

PR #127, opened 2015-07-03
Add interpolation settings to Targets

jcorral:
After some vegeta loving, after using it to test REST services, i got the idea that it wasn't designed to perform user behaviour load testing. 

With that on mind, I felt a little bit frustrated when i couldn't attack endpoints that were only attackable once, like identified database record creators.

So i came with the idea about giving the option to specify randomized deterministic or not deterministic interpolators for the Request URL and Body, to be able to modify the URL and Body on every hit.

By seeding a rand to setup the `Target`, i could even test a different endpoint afterwards with the same values.

As it have been so useful to me, i wanted to share it.

Ex:

```golang
seed := time.Now().UTC().Unix()
randomizer := rand.New(rand.NewSource(seed))

vegeta.NewStaticTargeter(&vegeta.Target{
		Method: "POST",
		URL:    "http://localhost:3002/foo/{interpolated_id}/bar",
		Body:   []byte(`{foo": "{interpolated_id}", "foobar": 1}`),
		URLInterpolators: []vegeta.URLInterpolator{
			&vegeta.RandomNumericInterpolation{
				Key:   "{interpolated_id}",
				Limit: int(^uint(0) >> 1),
				Rand:  randomizer,
			},
		},
		BodyInterpolators: []vegeta.BodyInterpolator{
			&vegeta.RandomNumericInterpolation{
				Key:   "{interpolated_id}",
				Limit: int(^uint(0) >> 1),
				Rand:  randomizer,
			},
		},
	})
```

### DIFF
--- a/lib/attack_test.go
+++ b/lib/attack_test.go
@@ -21,7 +21,7 @@ func TestAttackRate(t *testing.T) {
 	rate := uint64(100)
 	atk := NewAttacker()
 	var hits uint64
-	for range atk.Attack(tr, rate, 1*time.Second) {
+	for _ = range atk.Attack(tr, rate, 1*time.Second) {
 		hits++
 	}
 	if got, want := hits, rate; got != want {
--- a/lib/interpolations.go
+++ b/lib/interpolations.go
@@ -1,0 +1,47 @@
+package vegeta
+
+import (
+	"math/rand"
+	"strconv"
+	"strings"
+	"time"
+)
+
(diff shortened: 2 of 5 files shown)


tsenart:
Hi @jcorral,

Thanks for your contribution but I don't think there's any need for this functionality to be in vegeta itself.
You can easily create a script with generates URLs as you see fit and pipe them to vegeta with attack with the `-lazy` flag on.
```shell
$ ./genurls | vegeta attack -lazy ...
```

PR #130, opened 2015-08-08
Use R's 8th method of quantile estimation for latencies

### LINKED ISSUE #129: Quantiles are incorrectly calculated
In some executions of vegeta the metrics command generates incorrect quantiles, namely the value of the 99th quantile is the same as max, which is wrong. I double checked by dumping the results file into a CSV and calculating the quantiles using R. I'll be happy to provide the offending binary file (I see no way to attach it here).

tsenart:
This change implements R's 8th type of quantile estimation which is
approximately median-unbiased regardless of the data distribution.

Fixes #129  /cc @alexeyrodriguez

### DIFF
--- a/lib/metrics.go
+++ b/lib/metrics.go
@@ -1,50 +1,63 @@
 package vegeta
 
 import (
+	"math"
+	"sort"
 	"strconv"
 	"time"
-
-	"github.com/bmizerany/perks/quantile"
 )
 
-// Metrics holds the stats computed out of a slice of Results
-// that is used for some of the Reporters
-type Metrics struct {
-	Latencies struct {
+type (
+	// Metrics holds metrics computed out of a slice of Results which are used
+	// in some of the Reporters
+	Metrics struct {
+		// Latencies holds computed latency metrics.
+		Latencies LatencyMetrics `json:"latencies"`
+		// BytesIn holds computed incoming byte metrics.
+		BytesIn ByteMetrics `json:"bytes_in"`
+		// BytesOut holds computed outgoing byte metrics.
+		BytesOut ByteMetrics `json:"bytes_out"`
+		// Duration is the duration of the attack.
+		Duration time.Duration `json:"duration"`
+		// Wait is the extra time waiting for responses from targets.
+		Wait time.Duration `json:"wait"`
+		// Requests is the total number of requests executed.
+		Requests uint64 `json:"requests"`
+		// Rate is the rate of requests per second.
+		Rate float64 `json:"rate"`
+		// Success is the percentage of non-error responses.
+		Success float64 `json:"success"`
+		// StatusCodes is a histogram of the responses' status codes.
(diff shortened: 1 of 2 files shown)


tinnet:
:+1: noticed the same odd numbers that @alexeyrodriguez saw, came here, found the PR -> numbers now match what we thought they should be by looking at the raw data :rocket: 

:)

alexeyrodriguez:
I just tried out this PR, the numbers it reports for my test file are now correct.

:+1:

alexeyrodriguez:
The numbers match up to two decimal places, I'd expect they would be almost equal. FWIW, here is my R script:

  lat <- read.csv("results.csv")
  names(lat) <- c("time", "code", "latency", "out", "in", "err")
  lat$latency <- lat$latency / 1000000
  quantile(lat[lat$code==200,]$latency, probs = c(0.50,0.95,0.99,0.999,1))

tsenart:
@alexeyrodriguez: The default estimation method of R is `type = 7`. I implemented `type = 8`. Calculate it with: `quantile(lat[lat$code==200,]$latency, probs = c(0.50,0.95,0.99,0.999,1), type = 8)`

alexeyrodriguez:
Ah thanks, now it improved. I have numbers matching up to 4 decimal places, except for the 99th percentile which matches up to 3 decimal places.

tsenart:
@alexeyrodriguez: Can you send me your CSV? I'd like to investigate these discrepancies.

alexeyrodriguez:
I just sent the csv and bin files.

tsenart:
There were two issues with the R commands.

1. `read.csv` had `HEADER=true` which discarded the first line as a header, when it was data.
2. The results were filtered with `code==200` which doesn't match the semantics of `vegeta report`.

With those fixed, we almost get fully identical results. Now I'll have to dig into that `off-by-one` in the 95th percentile.

```console
$ cat results.bin | ./vegeta report -reporter=json | jq ".latencies"
{
  "mean": 2515792,
  "50th": 2032866,
  "95th": 5144141,
  "99th": 8917439,
  "max": 204889823
}
$ cat results.bin | ./vegeta dump -dumper=csv > results.csv
$ R --quiet
> lat <- read.csv("results.csv", header=FALSE)
> names(lat) <- c("time", "code", "latency", "out", "in", "err")
> quantile(lat$latency, probs = c(0.50,0.95,0.99,1), type = 8)
      50%       95%       99%      100%
  2032866   5144142   8917439 204889823
```

tsenart:
Alright, I found the bug, I had to round the `quantileR8` float64 results before converting them back to `time.Duration` which is an `int64`.

tsenart:
@alexeyrodriguez: Please test again, and let me know if everything matches up!

alexeyrodriguez:
I just tested it and it is an exact match! :100:

PR #139, opened 2015-09-06
Allow targets with custom bodies to be separated by empty line as stated in documentation

### LINKED ISSUE #138: Targets with custom bodies cannot be separated by empty line as stated in documentation
Target file with content like
```
POST http://foobar.org/fnord
Authorization: x12345
@vegeta-12345.txt

POST http://foobar.org/fnord/2
Authorization: y67890
@vegeta-67890.txt
```
leads to `bad target:` error.

coxx:
Fixes #138

### DIFF
--- a/lib/targets.go
+++ b/lib/targets.go
@@ -157,6 +157,7 @@ func NewLazyTargeter(src io.Reader, body []byte, hdr http.Header) Targeter {
 		if err = sc.Err(); err != nil {
 			return ErrNoTargets
 		}
+		_ = sc.Peek()
 		return nil
 	}
 }
--- a/lib/targets.go
+++ b/lib/targets.go
@@ -138,8 +138,22 @@ func NewLazyTargeter(src io.Reader, body []byte, hdr http.Header) Targeter {
 			if line = strings.TrimSpace(sc.Text()); line == "" {
 				break
 			} else if strings.HasPrefix(line, "@") {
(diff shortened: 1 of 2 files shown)


tsenart:
Hello @coxx,

Thanks for the bug report and patch! The second commit is its own feature and needs to be a separate PR. I'll review the first commit now.

lib/targets.go
```diff
@@ -157,6 +171,7 @@ func NewLazyTargeter(src io.Reader, body []byte, hdr http.Header) Targeter {
 		if err = sc.Err(); err != nil {
 			return ErrNoTargets
 		}
+		_ = sc.Peek()
```
tsenart: I think it would be a lot more reliable to consume all white space in the beginning of every invocation.
```go
		for sc.Scan() {
			if line := strings.TrimSpace(sc.Text()); line == "" {
				break
			}
		}
		tokens := strings.SplitN(line, " ", 2)
		if len(tokens) < 2 {
			return fmt.Errorf("bad target: %s", line)
		}
		switch tokens[0] {
		case "HEAD", "GET", "PUT", "POST", "PATCH", "OPTIONS", "DELETE":
```

That way we could also get rid of the `httpMethodChecker` regex.

tsenart:
@coxx: Ping.

coxx:
@tsenart 
Pong :-)
Sorry I started more complex refactoring but I was busy a bit.

Issue #141, opened 2015-09-07
Refactor target parser


tsenart:
The current parser code isn't very modular and could be simplified.

ei-grad:
@tsenart hi! Thanks for a great tool.

I want to deal with target parser refactoring to be able to use it in my own load testing tool. If there are nobody working on this currently, then you could expect a pull-request from me in a month or so (so long because my own load testing project is not the main thing I work on).

Would it be ok if I move target parsing to github.com/tsenart/vegeta/target package?

Any other suggestions / requirements / whishes?

tsenart:
Hello @ei-grad! Thanks for proposing to work on this. I suggest you approach this with small incremental steps and put all of this up for me to review regularly.

Let's start with everything in the same package as it currently is. Also please bring forward a bit more specifics of how you plan to refactor this.

tsenart:
Closed in favor of https://github.com/tsenart/vegeta/pull/300

PR #148, opened 2015-09-22
Allow targets contain inline bodies similar to shell "here tag"

coxx:
It allows targets to have bodies inline (instead of path to a file) as follows:

```
POST http://foo.bar/
Content-Type: application/json; charset=utf-8
@<<BODY
{
    "one": 1,
    "two": 2
}
BODY
```

Where BODY is arbitrary tag.

I understand that the patch does not add cleanliness to the target generator code but if idea and feature is OK I'll clean and refactor it in the next step.

### DIFF
--- a/lib/targets.go
+++ b/lib/targets.go
@@ -145,10 +145,24 @@ func NewLazyTargeter(src io.Reader, body []byte, hdr http.Header) Targeter {
 			if line = strings.TrimSpace(sc.Text()); line == "" {
 				break
 			} else if strings.HasPrefix(line, "@") {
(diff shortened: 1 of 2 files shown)


tsenart:
The idea is good, thanks! But I feel the refactor needs to happen before, not after this feature.

ifraixedes:
My understanding is that the refactor, you're talking about, refers to #141

missedone:
@tsenart , do you have any plan on the refactoring ticket #141, otherwise, looks like any target relate changes will be blocked until the refactoring done?

tsenart:
@missedone: Yeah, just didn't find time to do it.

tsenart:
Superseded by #260

Issue #153, opened 2015-10-29
Cannot install vegeta using go get


ram-nadella:
```sh
[~]> go get github.com/tsenart/vegeta
# github.com/tsenart/vegeta/lib
development/go/src/github.com/tsenart/vegeta/lib/results.go:58: syntax error: unexpected range, expecting {
development/go/src/github.com/tsenart/vegeta/lib/results.go:68: syntax error: unexpected }
```

```sh
[vegeta (master)]> go fmt lib/results.go
lib/results.go:58:7: expected operand, found 'range'
lib/results.go:61:4: expected '{', found 'if'
lib/results.go:62:5: expected operand, found 'continue'
lib/results.go:68:1: expected declaration, found '}'
exit status 2
```

tsenart:
You're Go version is probably too old.

ram-nadella:
I am running 1.3.3

```sh
[~]> go version
go version go1.3.3 linux/amd64
```

What is the minimum required version of go? Could you specify that in the README?

tsenart:
Recommended version is >= 1.5. Feel free to open a PR with the change to
the README.

tsenart:
But you should really be using releases: https://github.com/tsenart/vegeta/releases

PR #157, opened 2015-11-18
Mention go version requirement in the source section of README

ram-nadella:
Updated README as discussed in #153


README.md
```diff
@@ -24,6 +24,7 @@ command:
 $ go get github.com/tsenart/vegeta
```
tsenart: You can just change the initial paragraph to include the version constraint. **You need `go 1.5.2` installed (...)**

PR #160, opened 2015-11-27
attack: Implement -insecure flag

tsenart:
This commit adds a flag to ignore invalid server TLS certificates.

/cc @aidylewis


aidylewis:
This is brilliant news. Thank you so much. I look after my daughter at weekends, but can test this out and reply to your email when I am back at work on Monday.

PR #171, opened 2016-04-05
README: simplify Installing -> Source commands

slimsag:
- `go install github.com/tsenart/vegeta` is not needed because `go get` already performs an install.
- Add `-u` flag to `go get` to ensure that the latest version is installed.


slimsag:
Just something I noticed!

tsenart:
Thanks!

PR #177, opened 2016-04-20
RFC: WeightedTargetter

keegancsmith:
Is this something that you would be interested in merging upstream if developed further? (integrate into CLI, etc)

### DIFF
--- a/lib/internal/discreterand/alias.go
+++ b/lib/internal/discreterand/alias.go
@@ -1,0 +1,103 @@
+/*
+   Package discreterand provides an implementation of Vose's alias method for
+   choosing elements from a discrete distribution.
+
+   Copyright (c) 2013 Damian Gryski <damian@gryski.com>
+   Licensed under the MIT license.
+
+   For a full description of the algorithm, see
+   http://www.keithschwarz.com/darts-dice-coins/
+*/
+package discreterand
+
+import (
+	"math/rand"
+)
+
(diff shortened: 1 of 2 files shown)


tsenart:
Hey, thanks for starting this discussion. This looks quite useful, but there are challenges in bringing this to the cli tool. How would you define weights in the current target syntax? I don't immediately see a natural way to do it, and in order to bring this in, it'd have to be used in the cli too.

keegancsmith:
Yeah I agree, and I wasn't sure on any good syntax for it. But given I had this code I thought I'd kick off the discussion with a PR :)

Possible options I thought of initially:
* An optional weight at the end of the initial line. So your first line could be `GET http://user:password@goku:9090/path/to 5`
* A special header key value

The special header key value could be specified starting with something that is not allowed in usual headers, like indentation or maybe a =

Everything would have a default weight of 1. If no weights are specified we use the normal static targetter which is round-robin.

keegancsmith:
Note that this code can easily live outside of vegeta repo and only be for users who are using it as a lib. But if you are interested in it being a part of the cli I am willing to make it work.

tsenart:
@keegancsmith: Hm, I feel those suggestions are a bit hacky for the cli. I just wondered if this couldn't be achieved by repeating the targets proportionally to their desired weight. Like:

```
GET http://goku/foo
GET http://goku/foo

PR #195, opened 2016-08-22
Made headers in target definitions retain case sensitivity

### LINKED ISSUE #191: Vegeta modifies header values
I am using vegeta to load test, and the use case is to POST request with headers. I pass "X_MESSAGE_ID" but it is received at server as "X_message_id"

I understand that underscores are not used in headers as per RFC 822, but in my case I have to use them. Is there a flag to avoid this transformation ?

chriskirkland:
Fixes #191.  Based on solution in #124.

### DIFF
--- a/lib/targets.go
+++ b/lib/targets.go
@@ -159,7 +159,10 @@ func NewLazyTargeter(src io.Reader, body []byte, hdr http.Header) Targeter {
 					return fmt.Errorf("bad header: %s", line)
 				}
 			}
-			tgt.Header.Add(tokens[0], tokens[1])
+			// Add key/value directly to the http.Header (map[string][]string).
+			// http.Header.Add() canonicalizes keys but vegeta is used
+			// to test systems that require case-sensitive headers.
+			tgt.Header[tokens[0]] = append(tgt.Header[tokens[0]], tokens[1])
 		}
 		if err = sc.Err(); err != nil {
 			return ErrNoTargets


chriskirkland:
Simple tests:

### Vegeta
```
~$ cat targets.txt
GET http://localhost:8080
X-Account-ID: 8675309

GET http://localhost:8080
X-ACCOUNT-ID: 9035768

~ $ ./vegeta attack -duration=1s -rate=20 -targets=targets.txt | vegeta report
Requests      [total, rate]            20, 21.05
Duration      [total, attack, wait]    955.28554ms, 949.999921ms, 5.285619ms
Latencies     [mean, 50, 95, 99, max]  5.211109ms, 5.285619ms, 6.427649ms, 6.427649ms, 6.453087ms
Bytes In      [total, mean]            0, 0.00
Success       [ratio]                  0.00%
Status Codes  [code:count]             0:20
Error Set:
```

### Python simple HTTP server
<pre>
```
Listening on localhost:8080

----- Request Start ----->

/
Host: localhost:8080
User-Agent: Go-http-client/1.1
<b>X-Account-ID: 8675309</b>
Accept-Encoding: gzip

<----- Request End -----

127.0.0.1 - - [22/Aug/2016 10:10:58] "GET / HTTP/1.1" 200 -

----- Request Start ----->

/
Host: localhost:8080
User-Agent: Go-http-client/1.1
<b>X-ACCOUNT-ID: 9035768</b>
Accept-Encoding: gzip

<----- Request End -----

127.0.0.1 - - [22/Aug/2016 10:10:58] "GET / HTTP/1.1" 200 -

...
```
</pre>

tsenart:
Thanks!

PR #202, opened 2016-10-08
Doc fix: Results -> Result type elements

oneness:
Reading your code as I am picking up go. I like the overall design so far.


lib/results.go
```diff
@@ -27,7 +27,7 @@ type Result struct {
 // End returns the time at which a Result ended.
 func (r *Result) End() time.Time { return r.Timestamp.Add(r.Latency) }
 
-// Results is a slice of Results.
+// Results is a slice of Result type elements
```
tsenart: Please finish the sentence with a `.`

oneness: oops, just did. :)

PR #213, opened 2017-01-20
WIP add support for passing allowed custom methods

topliceanu:
See #214


topliceanu:
Hi! I'm in a bit of a conundrum and I need some help. I forked this repo on our org but I had to change the import paths from `github.com/tsenart/vegeta` to `github.com/pusher/vegeta` so tests would work. They don't work here though! What is the correct way to go over this?

Issue #214, opened 2017-01-20
Proposal: support custom http verbs


topliceanu:
### rationale
- http/2 allows using custom http verbs (and I think they were allowed in http/1.1 as well) and there's no reason why vegeta can't support them.
- while not strictly RESTful, custom verbs have better semantics when working with resources.

### context 
- I ran `echo "SUBSCRIBE http://localhost" | vegeta attack -http2 | tee /tmp/results.vegeta | vegeta report` and I got back `2017/01/19 21:58:42 bad method: SUBSCRIBE` and an empty report.

### implementation
- ~~`vegeta attack` gets a new options `-allowed-methos=APPEND,SUBSCRIBE` which is a list of comma-separated verbs allowed in targets. This list of allowed methods gets passed to the Targeter~~ `vegeta attack` allows any ASCII upper-case word to be used as HTTP verb.
- I also added a test to make sure Targets are correctly configured
- ~~a WIP PR is here  https://github.com/tsenart/vegeta/pull/213~~ The new PR is here https://github.com/tsenart/vegeta/pull/217

tsenart:
I'd just allow any string as method instead of defining the allowed ones per attack.

topliceanu:
My thinking was that the targets file could come from a source you don't control, so you might need some method of checking. However you are absolutely right, allowing any word to be a verb is a MUCH simpler solution. I will open a PR for that.

PR #217, opened 2017-01-28
PTAL allow any upper-case ASCII word to be an http method

topliceanu:
The motivation and context for this PR are detailed in the proposal issue: https://github.com/tsenart/vegeta/issues/214

The implementation has changes as per maintainer's suggestions:
- with this PR vegeta now supports any upper-case ASCII word as a valid HTTP verb, not just  POST/GET/PUT/PATCH/DETELE/HEAD/OPTIONS...

CC @tsenart

### DIFF
--- a/lib/targets.go
+++ b/lib/targets.go
@@ -127,12 +127,10 @@ func NewLazyTargeter(src io.Reader, body []byte, hdr http.Header) Targeter {
 		if len(tokens) < 2 {
 			return fmt.Errorf("bad target: %s", line)
 		}
-		switch tokens[0] {
-		case "HEAD", "GET", "PUT", "POST", "PATCH", "OPTIONS", "DELETE":
-			tgt.Method = tokens[0]
-		default:
+		if !isHttpMethod(tokens[0]) {
 			return fmt.Errorf("bad method: %s", tokens[0])
 		}
+		tgt.Method = tokens[0]
 		if _, err = url.ParseRequestURI(tokens[1]); err != nil {
 			return fmt.Errorf("bad URL: %s", tokens[1])
 		}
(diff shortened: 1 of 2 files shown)


lib/targets.go
```diff
@@ -171,12 +169,18 @@ func NewLazyTargeter(src io.Reader, body []byte, hdr http.Header) Targeter {
 	}
 }
 
-var httpMethodChecker = regexp.MustCompile("^(HEAD|GET|PUT|POST|PATCH|OPTIONS|DELETE) ")
+var httpMethodChecker = regexp.MustCompile("^[A-Z]*$")
 
-func startsWithHTTPMethod(t string) bool {
+func isHttpMethod(t string) bool {
```
tsenart: Why can't we have just `startsWithHTTPMethod` as it was before with the regex `^[A-Z]+`?

topliceanu: That is a good point. Updated!

lib/targets.go
```diff
@@ -171,8 +169,9 @@ func NewLazyTargeter(src io.Reader, body []byte, hdr http.Header) Targeter {
 	}
 }
 
-var httpMethodChecker = regexp.MustCompile("^(HEAD|GET|PUT|POST|PATCH|OPTIONS|DELETE) ")
+var httpMethodChecker = regexp.MustCompile("^[A-Z]+\\s")
 
+// a line starts with an http method when the first word is uppercase ascii followed by an url
```
tsenart: This comment doesn't really capture what the function does.

tsenart:
Thanks!

topliceanu:
Hi @tsenart, do you want to do a new [minor release](https://github.com/tsenart/vegeta/releases) which contains this change? Not sure what the release process is?

tsenart:
@topliceanu: Done.

topliceanu:
Thank you

PR #220, opened 2017-02-25
Swapped out Vegeta img for more accurate GIF

wgottschalk:
Saturday morning PR here. I feel like this GIF more accurately depicts what the Vegeta library is all about!


tsenart:
Ahah, thanks but no thanks :-) Too distracting.

PR #222, opened 2017-03-03
mark responses as success in no redirect following mode.

keijiyoshida:
## Issue which this PR solves

Vegeta does not mark responses as success when `-1` is set to the attack `-redirects` option. `Bytes In`, `Bytes Out`, `Success` and `Status Codes` of a report are zeros as below:

```shell
echo "GET http://xxxx.xxxx" | vegeta attack -redirects -1 -rate 1 -duration 10s | tee results.bin | vegeta report
Requests      [total, rate]            10, 1.11
Duration      [total, attack, wait]    9.003078071s, 8.999999898s, 3.078173ms
Latencies     [mean, 50, 95, 99, max]  12.427662ms, 3.246776ms, 3.823605ms, 3.823605ms, 96.816248ms
Bytes In      [total, mean]            0, 0.00
Success       [ratio]                  0.00%
Status Codes  [code:count]             0:10
Error Set:
```

## Changes in this PR

Have `a.client.CheckRedirect` return `http.ErrUseLastResponse` when `-1` is set to the attack `-redirects` option. By doing this, Go's HTTP client does not follow redirects and returns the first HTTP response with its body unclosed. See https://golang.org/src/net/http/client.go?h=ErrUseLastResponse#L398 for the details.

### DIFF
--- a/lib/attack.go
+++ b/lib/attack.go
@@ -7,7 +7,6 @@ import (
 	"io/ioutil"
 	"net"
 	"net/http"
-	"strings"
 	"sync"
 	"time"
 
@@ -95,10 +94,14 @@ func Redirects(n int) func(*Attacker) {
 	return func(a *Attacker) {
 		a.redirects = n
 		a.client.CheckRedirect = func(_ *http.Request, via []*http.Request) error {
-			if len(via) > n {
+			switch {
+			case n == NoFollow:
+				return http.ErrUseLastResponse
+			case n < len(via):
 				return fmt.Errorf("stopped after %d redirects", n)
+			default:
+				return nil
 			}
-			return nil
 		}
 	}
 }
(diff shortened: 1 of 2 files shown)


lib/attack.go
```diff
@@ -94,11 +93,17 @@ func Connections(n int) func(*Attacker) {
 func Redirects(n int) func(*Attacker) {
 	return func(a *Attacker) {
 		a.redirects = n
-		a.client.CheckRedirect = func(_ *http.Request, via []*http.Request) error {
-			if len(via) > n {
-				return fmt.Errorf("stopped after %d redirects", n)
+		if n == NoFollow {
```
tsenart: I'd rather have this code be the simpler version:

```go
a.client.CheckRedirect = func(_ *http.Request, via []*http.Request) error {
    switch {
    case n == NoFollow:
        return http.ErrUseLastResponse
    case n < len(via):
        return fmt.Errorf("stopped after %d redirects", n)
    default:
        return nil
    }
}
```

tsenart:
Can you please add a regression test that fails for the bug you're fixing?

keijiyoshida:
@tsenart Thanks for your reviewing and sorry for the late response. I fixed the module and added a regression test. Thanks.

PR #229, opened 2017-04-26
Add ability to add context to a Target.

chrismoos:
By adding context a multi-stage attack can be carried out. The context for the `Target` is passed through to `Result` which allows you to execute another attack based on the additional context
inside the `Result`.

For example, if you want to first create a session in the first HTTP request and then use that session ID in the second request:

```go
for res := range loginAttacker.Attack(targeter, rate, duration) {
    sessionIdChan <- res.Context["sessionId"].(string)
}
```

Then the other `Targeter` can read the `sessionId` for the next request in the stage:

```go
return func(tgt *vegeta.Target) error {
    sessionId := <- sessionIdChan
}
```


tsenart:
Hello,

I don't want to introduce a free-form context map into the `Result` and `Target` types in the library without a symmetric CLI use case and interface. Doing so would violate the design consistency we now have.

However, note that we already have an `http.Header` in the `Target` type, and as this is exposed through the CLI as well, it'd be rather simple to pass it back in the `Result` type.

What do you think? Would this solve your particular problem?

chrismoos:
Thanks for the response  @tsenart.

It's possible but those headers would end up getting sent into the HTTP request and also limits what  kind of context you can pass through (i.e just a string).

I've actually used the context to implement a generic, multi-stage _attack pipeline_. It might give you a better idea of what I'm trying to do. See https://gist.github.com/chrismoos/0673df867b1f9e820c5f2bab1e6f1af5. 

Also, I think a complex multi-stage attack (that has contextual stages) would be difficult just via CLI...probably there would need to be some sort of scripting to facilitate that.

Sample usage of the above multi-stage attack:

```golang
pipeline := NewAttackPipeline()
pipeline.Add(stage1Targeter, resultHandler, rate, duration)
pipeline.Add(stage2Targeter, resultHandler, rate, duration) 
pipeline.Run(vegeta.Timeout(10*time.Minute), vegeta.KeepAlive(true))
```

tsenart:
That's a cool use-case! Let's work through this then. What sort of context data do you imagine wouldn't fit nicely in HTTP headers? I'm adamant we keep the consistency between lib and cli. If we were to extend the `Target` with a general context map, we'd need a way to represent it in the plaintext target language schema that feeds into `vegeta attack`.

I don't see how to do it cleanly with the current language, which is quite close to plain old HTTP 1.1

chrismoos:
The usefulness of the context is mostly derived by extracting something from the HTTP response (i.e a response header, something from body, etc,.) and then passing it through the context to the next stage of the attack.

If you want to integrate this into the current target language schema there would need to be a facility to run a script/handler/etc,. when evaluating the target. One way to do it might be to allow for a user-defined script to be set for a target.

For example:

```http
POST http://goku:9090/things
X-Account-ID: 99
@/path/to/newthing.json;post-run=/path/to/post-run.sh
```
`/path/to/post-run.sh`:
```bash
echo '{"variableInContext": "test123"}'
```

The `post-run` script would be invoked and the full HTTP response (raw, or maybe parsed and placed into a JSON object, for example) sent into `stdin`. The script would have a defined set of exit codes to indicate success and failure(s). In addition, the script can write to `stdout` a JSON object that contains the context associated with the request. 

Then, to be able to use the context, we can use variable interpolation (or full featured `text/template` support) to allow for a target to use a context variable:

```http
PATCH http://goku:9090/thing/71988591
My-Customer-Header: ${variableInContext}
@/path/to/thing-71988591.json
```

tsenart:
Not pursuing this approach for now.

kiwiidb:
Sorry for the threadomancy but I'd like to know if I could use Vegeta for my use case using the approaches described here. I'd like to make alternate calls creating a resource and then delete it in the next call, using a UID parsed from the response body of the first call. (How) can I do this using vegeta?

PR #234, opened 2017-05-17
Add option to override HTTP Proxy on Attacker.

chrismoos:
Currently an `Attacker` sets the `Proxy` field to `http.ProxyFromEnvironment` which uses a proxy if the environment variables are present. This means there is no way to use a proxy only for a _specific_ attacker -- it's global. 

This change adds the ability to override the proxy settings on an `Attacker` instance.

### DIFF
--- a/lib/attack.go
+++ b/lib/attack.go
@@ -7,6 +7,7 @@ import (
 	"io/ioutil"
 	"net"
 	"net/http"
+	"net/url"
 	"sync"
 	"time"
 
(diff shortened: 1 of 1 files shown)


tsenart:
Can you add a test please?

tsenart:
OK, I'll merge this and add a test. Thanks!

marcoscoelho:
There a option to use it via command line ?

tsenart:
You can set the `HTTP_PROXY` environment variable.

Issue #235, opened 2017-05-29
Real like traffic test


stereobooster:
Thanks, for cool project.

I have requests breakdown from production system (getting this data from ELK).

It looks something like:

- request 1 - 50%
- request 2 - 20%
- request 3 - 10%
- request 4 - ...

So i want to script the same type of requests and make vegeta to do "real-like traffic" so it preserves distribution of requests. 

Then I want it slowly start to increase number of requests until service starts to fail. E.g. error rate get higher some level.

Is it possible?

**UPD**

So I was thinking of something like this:

```
POST http://goku:9090/things
@/path/to/newthing.json
%60

PATCH http://goku:9090/thing/71988591
@/path/to/thing-71988591.json
%40
```

thalesfsp:
@stereobooster Have your changes been merged? @tsenart

stereobooster:
no

thalesfsp:
@stereobooster What’s missing so we can get a PR? Do you need help? Really cool and welcome the feature that you described!

stereobooster:
https://github.com/tsenart/vegeta/pull/236#issuecomment-328530817

PR #236, opened 2017-05-30
[#235] Naive implementation of percentage distribution

stereobooster:
I need to add more tests. And the code can be not that idiomatic (I'm not using go that much). Give me some feedback and I will finish it

Usage (targets file):

```
POST http://goku:9090/things
@/path/to/newthing.json
%60

PATCH http://goku:9090/thing/71988591
@/path/to/thing-71988591.json
%40
```

### DIFF
--- a/lib/targets.go
+++ b/lib/targets.go
@@ -13,6 +13,9 @@ import (
 	"strings"
 	"sync"
 	"sync/atomic"
+	"strconv"
+	"sort"
+	"math/rand"
 )
 
 // Target is an HTTP request blueprint.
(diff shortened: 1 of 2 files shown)


tsenart:
This can be easily accomplished by running two concurrent attacks with different targets and parameters. It's not worth the additional complexity.

stereobooster:
> two concurrent attacks

or 10 if you have 10 targets ¯\_(ツ)_/¯

tsenart:
Maybe this helps: https://github.com/bbc/FOSS-Load-Tools/blob/master/vegeta/vegeta.md

PR #260, opened 2017-09-18
Support targets with inline bodies similar to shell's "here tag"

1u0:
Hi, this is an updated PR of the effort done by @coxx in #148.

The change will allow `vegeta` to read body payloads from the same target data stream.


1u0:
@tsenart please consider this functionality, maybe not as done in this PR but at least in general.

The use case is when there are different request bodies in the payloads. It would be convenient to have payload in one file (or generated dynamically and passed through stdin) instead of having a lot of scattered files.

mbrookes1304:
Hi I'm interested to know what is the status of this?

mbrookes1304:
I see the commits are from quite some time ago, I'm happy doing the work to rebase etc for merging into `master`

tsenart:
@mbrookes1304: I'm reviewing the implementation and thinking through the design now.

tsenart:
Superseded by #300.

PR #261, opened 2017-09-19
Addedd support for h2c requests (HTTP/2 without TLS)

mzdeb:
Hi,

It's my first PR here, and first in golang. ;) Thank you very much for your hard work on this project.

Probably support for h2c might be done better, so any suggestions welcome.

### DIFF
--- a/attack.go
+++ b/attack.go
@@ -30,6 +30,7 @@ func attackCmd() command {
 	fs.StringVar(&opts.keyf, "key", "", "TLS client PEM encoded private key file")
 	fs.Var(&opts.rootCerts, "root-certs", "TLS root certificate files (comma separated list)")
 	fs.BoolVar(&opts.http2, "http2", true, "Send HTTP/2 requests when supported by the server")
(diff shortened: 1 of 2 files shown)


tsenart:
How does this relate to https://github.com/golang/go/issues/14141?

mzdeb:
As bradfitz wrote:
> I've already accepted a ton of logic for h2c. My policy has been and remains: it should be possible for people to do h2c via a separate package, but it won't be supported by default.
> All the pieces should be there now for you to glue it together yourself.
> For server-side support you'll need to hijack in the PRI * HTTP/2.0 Handler and then call http2.Server.ServerConn.

So currently H2C is quite simple for client to implement. When golang/go#16581 will be resolved, all options from ```http.Transport``` will be also available.

awinder:
I've gotta 👍 this issue because I'm currently impacted by the same issue and it looks like several months have elapsed without any closure from the upstream branch.  Any chance of reviving this?

PR #136, opened 2015-09-04
Streaming reporters and dumpers

tsenart:
This changeset makes vegeta reporters and dumpers work with bounded
memory by replacing batch algorithms with streaming counterparts.

Fixes #132

Issue #132, opened 2015-08-15
Report processing runs out of memory with large report file.


ohookins:
Report file was 1.5GB, physical memory on this instance is about 7GB. Anything I can do to successfully process this report?

tsenart:
Hi Oliver,

The reporters are currently implemented to process in batch. That explains your OOM. Which reporter were you using? What were the parameters to the attack stage?

ohookins:
I used the default parameters to report, so I guess `text`.
On the attack it was this: `cat list_of_targets | vegeta attack -duration=24h -lazy=true -rate=200 > results.bin`
The list of targets was a file with 13.5 million URLs.

tsenart:
Wow. That's some impressive scale, my friend :-) What's the use case for a 24h load test? I'm very curious!

ohookins:
Actually I'm just looking to hit the servers with all of the requests in the input file to pre-warm some caches. The 24h period and request rate should be enough to cover all of the target URLs in the file, which is the only reason behind those settings - ideally I'd ask the test tool to just cycle sequentially through the targets until complete, but that's not an option.

tsenart:
@ohookins: From your explanation, you're not actually looking to use vegeta's reports, only the load generation with `attack`. In any case, I'm working on making the reporters work with in streaming, but won't be able to finish it too soon. However, I got something that should work for you to process that data set with report. Would you mind building the branch https://github.com/tsenart/vegeta/tree/streaming-reporters and running:
```console
$ cat results.bin | ./vegeta report -window=24h
```

Let me know if it worked.

ohookins:
Yes, that did work.

Actually I do need to use the reports, since I need to tell what level of success the load test had. For example I didn't set high enough ulimit and I think that led to a bunch of problems resolving DNS and creating new sockets. It's also still useful to see if there were any significant performance problems during these kinds of tests to judge whether it has been successful or just hitting something which doesn't perform any computation.

tsenart:
OK, makes sense. Remember though, that you can also use the dump command in combination with jq to analyse the raw results. For your case, something like this would work:
```console
$ cat results.bin | vegeta dump | jq '.error'
```

ohookins:
:cool:

PR #268, opened 2017-10-27
Add 90th percentile to the results

daluu:
Quite a few tools offer 90th percentile statistic, and that's one metric we do monitor. e.g. jmeter, wrk2, etc. I don't see why vegeta shouldn't include it. Right now 95th % is the closest thing to it, but still that's a higher number.

I assume I got it reasonable but please review the quantile ratios setup and the expected latencies for each quantile with this addition. Which leads to the below question:

And on a related note, any reason to use the specific quantiles package instead of writing code from scratch or using another statistical/percentile/quantile library? I came across these in a quick search, not sure how they compare against the current library used.

https://github.com/montanaflynn/stats/blob/master/percentile.go
https://github.com/bmizerany/perks/tree/master/quantile

I'm not a statistician, so can't say much on this but the current quantile library used at least is a bit confusing to the novice user since we are free to define the ratio's for each percentile - one could make mistakes. Contrast that with percentile libraries for python, node.js/javascript, etc. and it's more straightforward: import library and call a percentile method against an array of values, specifying the desired percentile, no mucking with some percentile/quantile ratio value that maps it to the percentile accuracy.

### DIFF
--- a/README.md
+++ b/README.md
@@ -281,13 +281,13 @@ Specifies the kind of report to be generated. It defaults to text.
 
 ##### `text`
 ```console
-Requests      [total, rate]             1200, 120.00
-Duration      [total, attack, wait]     10.094965987s, 9.949883921s, 145.082066ms
-Latencies     [mean, 50, 95, 99, max]   113.172398ms, 108.272568ms, 140.18235ms, 247.771566ms, 264.815246ms
-Bytes In      [total, mean]             3714690, 3095.57
-Bytes Out     [total, mean]             0, 0.00
-Success       [ratio]                   55.42%
-Status Codes  [code:count]              0:535  200:665
+Requests      [total, rate]                1200, 120.10
+Duration      [total, attack, wait]        9.992530625s, 9.991666185s, 864.44µs
+Latencies     [mean, 50, 90, 95, 99, max]  1.623143ms, 1.420976ms, 2.808559ms, 2.941666ms, 3.120381ms, 4.092141ms
+Bytes In      [total, mean]                394800, 329.00
+Bytes Out     [total, mean]                0, 0.00
+Success       [ratio]                      55.42%
+Status Codes  [code:count]                 0:535  200:665
 Error Set:
 Get http://localhost:6060: dial tcp 127.0.0.1:6060: connection refused
 Get http://localhost:6060: read tcp 127.0.0.1:6060: connection reset by peer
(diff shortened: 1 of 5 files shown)


lib/metrics.go
```diff
@@ -140,8 +143,9 @@ func (m *Metrics) init() {
 	if m.latencies == nil {
 		m.latencies = quantile.New(
 			quantile.Known(0.50, 0.01),
-			quantile.Known(0.95, 0.001),
-			quantile.Known(0.99, 0.0005),
+			quantile.Known(0.90, 0.001),
```
tsenart: What was the thought process behind these tolerance values?

daluu: See my original PR comment for background to this, I did ask for review of the implementation. But to answer your question, naive gut instinct, based on the existing implementation. I don't know if it's correct, but seems close enough? This kind of implementation to a math novice is less straightforward than typical statistical percentile libraries. What should be the "proper" values? And while we're at this, how to calculate the tolerance values for any given percentile in case other users might want something else like 75-80th percentiles (relative to the current ones).

tsenart:
OK, I'm cool with this. Can you please add a test for the changes and update the documentation (README, etc) with the new field? Thanks.

daluu:
Ok, I'll update the docs. For test, I already made change in initial commit. Is that test update good enough?

tsenart:
> And on a related note, any reason to use the specific quantiles package instead of writing code from scratch or using another statistical/percentile/quantile library? I came across these in a quick search, not sure how they compare against the current library used.

Here's the context for this choice. In sum, by using streaming (online) algorithms, we have an upper bound on resource usage.

https://github.com/tsenart/vegeta/issues/132
https://github.com/tsenart/vegeta/pull/136

With that said, I sympathise with your trouble in understanding which error tolerances are appropriate for each known quantile. In the past, I don't think I have put much thought in the choice of those values, falling back instead to the then documented defaults.

It's perhaps a good time to evaluate and benchmark these libraries as alternatives. The first implements the same algorithm but doesn't require the user to define the error tolerances explicitly. The other two are entirely different algorithms.

- https://github.com/bmizerany/perks/tree/master/quantile
- https://github.com/dgryski/go-gk
- https://github.com/spenczar/tdigest

daluu:
Thanks for the insight/clarification @tsenart, I'ved filed an issue with the original quantile library for clarification on determining the default tolerances for those common percentiles.

daluu:
@tsenart for the docs, what specific test profile/configuration did you run with to get the sample results presented (latency, duration, bytes I/O)? So that I can update adding in the 90th percentile latency that is a close match to the original results. Or do you not care that I run any new vegeta test to get sample results to overwrite with in order to include the new 90th percentile?

tsenart:
The tests look fine to me as they are. I was wondering if you will have time to evaluate the other quantile libraries. If not, I'll take a stab at it.

daluu:
I might have some time next month to evaluate the other libraries, and it would be good idea for you to take a stab at it regardless.

tsenart:
@daluu: It's likely we're migrating to a new library. Check this branch out: https://github.com/tsenart/vegeta/tree/evaluate-quantile-libs

tsenart:
Hey @daluu, we have a new quantile estimation library in master. Feel free to adapt your code to it.

tsenart:
@daluu: That’s a third party library. Make sure to install the dependencies using make vendor

tsenart:
However! We don’t need to make that public at all. The users of the library already have access to the raw Results which contain the latencies. So they can use an external estimator library if they do wish. I don’t want to give this optionality and added complexity if it’s not warranted.

README.md
```diff
@@ -301,33 +301,34 @@ Get http://localhost:6060: http: can't write HTTP request on broken connection
...
+  "wait": 864440,
+  "requests": 1200,
+  "rate": 120.10008919248136,
   "success": 1,
   "status_codes": {
-    "200": 100
+    "200": 1200
   },
-  "errors": []
+  "errors": null
```
tsenart: I find it strange this became `null` from `[]`. Can you track this down?

README.md
```diff
@@ -406,12 +407,12 @@ It'll read and sort them by timestamp before generating reports.
...
 ```console
 $ vegeta report -inputs="10.0.1.1.bin,10.0.2.1.bin,10.0.3.1.bin"
-Requests      [total, rate]         3600000, 60000.00
-Latencies     [mean, 95, 99, max]   223.340085ms, 326.913687ms, 416.537743ms, 7.788103259s
-Bytes In      [total, mean]         3714690, 3095.57
-Bytes Out     [total, mean]         0, 0.00
-Success       [ratio]               100.0%
-Status Codes  [code:count]          200:3600000
+Requests      [total, rate]             3600000, 60000.00
+Latencies     [mean, 90, 95, 99, max]   223.340085ms, 318.808559ms, 326.913687ms, 416.537743ms, 7.788103259s
```
tsenart: Do you feel the report is still readable with this addition?

daluu: Yes, IMHO I think it's still readable (except have to check on Windows, where terminal window width isn't much resizable). Without the addition, the line spans more than the conventional 80 characters line width anyhow.

What could help with or without this addition is to reduce the significant digits after the decimal in the latencies (as well as other fields). Do we really need that many digits? 2 to 3 after decimal should suffice, and that will significantly reduce line width.

tsenart:
Ping :-)

daluu:
Thanks for the ping, kinda busy lately, haven't gotten time to finish against the updates from master. I'll resume when I get a chance soon.

tsenart:
Closing this until further work.

PR #270, opened 2017-11-28
feat(gzip): add the possibility to compress the payload

FlorentinDUBOIS:


lib/attack.go
```diff
@@ -237,7 +243,7 @@ func (a *Attacker) hit(tr Targeter, tm time.Time) *Result {
 		return &res
 	}
 
-	req, err := tgt.Request()
+	req, err := tgt.Request(a.gzip)
```
tsenart: Is there an advantage of performing the compression on every request? If not, I'd suggest to do it once and store the result in the `Target.Body`.

attack.go
```diff
@@ -107,6 +111,17 @@ func attack(opts *attackOpts) (err error) {
 		src = files[opts.targetsf]
 		hdr = opts.headers.Header
 	)
+	if opts.gzip {
+		buffer := new(bytes.Buffer)
+		w := gzip.NewWriter(buffer)
+
+		hdr.Set("Content-Type", "application/gzip")
```
tsenart: We should be setting the `Content-Encoding header`: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Encoding

attack.go
```diff
@@ -107,6 +111,17 @@ func attack(opts *attackOpts) (err error) {
...
+
+		w.Write(body)
+		w.Close()
```
tsenart: Please handle the errors being returned.

PR #280, opened 2018-03-22
Add jplot integration usage example to the README

### LINKED ISSUE #252: Plotly.js for real-time analysis
We know we can loop against the report to generate real-time stats. What do you think about sed/awk and posting to a service with Plotly?

rs:
Fixes #252


tsenart:
Thanks!

PR #282, opened 2018-04-26
Add support for custom DialContext on http.Transport

mpuncel:
This supports uses cases such as load testing a unix socket.

The implementation is a bit odd because there are attacker options for
tweaking settings on the standard *net.Dialer, however setting a custom
DialContext means those settings will be ignored. Conversely, setting
*net.Dialer options will cause the custom DialContext to be overwritten.

### DIFF
--- a/lib/attack.go
+++ b/lib/attack.go
@@ -1,6 +1,7 @@
 package vegeta
 
 import (
+	"context"
 	"crypto/tls"
 	"fmt"
 	"io"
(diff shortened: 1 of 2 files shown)


mpuncel:
To reveal some motivation for this change, I'm looking into load testing an [Envoy](https://github.com/envoyproxy/envoy) ("service mesh") setup, where I have microservices proxy outbound traffic through a sidecar Envoy over a unix socket.

The HTTP stack in go requires the scheme to be "http" or "https", and as I understand always uses the TCP stack, so you have to specify a custom DialContext. There wasn't a way to do this in vegeta, so I attempted to add it.

The implementation is a bit clumsy because there are some options which tweak settings on a *net.Dialer and then set the dial function on the transport from that *net.Dialer, which will overwrite the custom DialContext.

tsenart:
Vegeta was written before `context.Context` became a thing. Could there be a change that would satisfy your needs but would integrate `context.Context` in a more structured approach?

I imagine we could change the defaults sanely from a simple `Dial` to a `DialContext` and also expose a functional option to set the parent context of each request.

In regards to your particular need, would you please share a code snippet of what you're trying to do with the unix domain sockets? Thanks!

Issue #298, opened 2018-06-27
Inline post request body in targets file


mbrookes1304:
#### Proposal
Currently custom POST bodies are specified by referencing a separate file which contains the body to be sent. Instead it should be possible to specify the POST body in the file alongside the target

#### Background
We are load testing a service by injecting actual requests collected from the logs. Each request has a different body and because we are using 1000s of requests we do not want to have 1000s of files, one for each body.

#### Workarounds
We created a wrapper using the library that would read the list of requests & construct targets before launching the attack.

tsenart:
See #260.

PR #300, opened 2018-06-29
JSON targets format

tsenart:
#### Background

The original targets format closely resembles plain HTTP/1.x messages. It was meant to be used by humans writing those targets directly on their terminals. With its lack of support for in-line body definitions, this ill-defined legacy format made Vegeta hard to integrate with other programs generating dynamic targets on the fly with distinct request bodies.

Extending the legacy format to support arbitrary in-line request bodies would require further deviation from the HTTP/1.x message format since, to the best of my knowledge, there is no official plain text framing format that would allow the boundaries between multiple targets with arbitrary request bodies to be easily recognized (and hence, parsed).

It could be done, certainly, but for the use case of integrating Vegeta with other programs generating targets, I think a simpler alternative JSON based format could be better.

This PR proposes just such a format and is intended to gather feedback from interested parties.

It's currently a work-in-progress, but the CLI interface is already specified and documented.

Please provide feedback: @coxx, @1u0, @mbrookes1304, @cwinters, @xla.

##### Related Issues and PRs

- https://github.com/tsenart/vegeta/pull/148
- https://github.com/tsenart/vegeta/pull/260
- https://github.com/tsenart/vegeta/issues/298

#### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,6 +47,10 @@ attack command:
       Max open idle connections per target host (default 10000)
   -duration duration
       Duration of the test [0 = forever]
+  -format string
+      Targets format [legacy] (default "legacy")
+  -h2c
+      Send HTTP/2 requests without TLS encryption
   -header value
       Request header
   -http2
@@ -113,47 +117,7 @@ Specifies which profiler to enable during execution. Both *cpu* and
 #### `-version`
 Prints the version and exits.
 
-### `attack`
-```console
-$ vegeta attack -h
-Usage of vegeta attack:
-  -body string
-      Requests body file
-  -cert string
-      TLS client PEM encoded certificate file
-  -connections int
-      Max open idle connections per target host (default 10000)
-  -duration duration
-      Duration of the test [0 = forever]
-  -header value
-      Request header
-  -http2
-      Send HTTP/2 requests when supported by the server (default true)
-  -insecure
-      Ignore invalid server TLS certificates
-  -keepalive
-      Use persistent connections (default true)
-  -key string
-      TLS client PEM encoded private key file
-  -laddr value
-      Local IP address (default 0.0.0.0)
-  -lazy
-      Read targets lazily
-  -output string
-      Output file (default "stdout")
-  -rate uint
-      Requests per second (default 50)
-  -redirects int
-      Number of redirects to follow. -1 will not follow but marks as success (default 10)
-  -root-certs value
-      TLS root certificate files (comma separated list)
-  -targets string
-      Targets file (default "stdin")
-  -timeout duration
-      Requests timeout (default 30s)
-  -workers uint
-      Initial number of workers (default 10)
-```
+### `attack` command
 
 #### `-body`
 Specifies the file whose content will be set as the body of every
@@ -171,6 +135,56 @@ Specifies the amount of time to issue request to the targets.
 The internal concurrency structure's setup has this value as a variable.
 The actual run time of the test can be longer than specified due to the
 responses delay. Use 0 for an infinite attack.
+
+#### `-format`
+Specifies the targets format to decode.
+
+##### `legacy` format
+
+The ill-defined legacy format almost resembles the plain-text HTTP message format but
+doesn't support in-line HTTP bodies, only references to files that are loaded and used
+as request bodies (as exemplified below).
+
+Although targets in this format can be produced by other programs, it was originally
+meant to be used by people writing targets by hand for simple use cases.
+
+Here are a few examples of valid targets files in the legacy format:
+
+Simple targets
+```
+GET http://goku:9090/path/to/dragon?item=ball
+GET http://user:password@goku:9090/path/to
+HEAD http://goku:9090/path/to/success
+```
+
+Targets with custom headers
+```
+GET http://user:password@goku:9090/path/to
+X-Account-ID: 8675309
+
+DELETE http://goku:9090/path/to/remove
+Confirmation-Token: 90215
+Authorization: Token DEADBEEF
+```
+
+Targets with custom bodies
+```
+POST http://goku:9090/things
+@/path/to/newthing.json
+
+PATCH http://goku:9090/thing/71988591
+@/path/to/thing-71988591.json
+```
+
+Targets with custom bodies and headers
+```
+POST http://goku:9090/things
+X-Account-ID: 99
+@/path/to/newthing.json
+```
+
+#### `-h2c`
+Specifies that HTTP2 requests are to be sent over TCP without TLS encryption.
 
 #### `-header`
 Specifies a request header to be used in all targets defined, see `-targets`.
@@ -220,39 +234,6 @@ list. If unspecified, the default system CAs certificates will be used.
 Specifies the attack targets in a line separated file, defaulting to stdin.
 The format should be as follows, combining any or all of the following:
 
-Simple targets
-```
-GET http://goku:9090/path/to/dragon?item=balls
-GET http://user:password@goku:9090/path/to
-HEAD http://goku:9090/path/to/success
-```
-
-Targets with custom headers
-```
-GET http://user:password@goku:9090/path/to
-X-Account-ID: 8675309
-
-DELETE http://goku:9090/path/to/remove
-Confirmation-Token: 90215
-Authorization: Token DEADBEEF
-```
-
-Targets with custom bodies
-```
-POST http://goku:9090/things
-@/path/to/newthing.json
-
-PATCH http://goku:9090/thing/71988591
-@/path/to/thing-71988591.json
-```
-
-Targets with custom bodies and headers
-```
-POST http://goku:9090/things
-X-Account-ID: 99
-@/path/to/newthing.json
-```
-
 #### `-timeout`
 Specifies the timeout for each request. The default is 0 which disables
 timeouts.
@@ -262,17 +243,7 @@ Specifies the initial number of workers used in the attack. The actual
 number of workers will increase if necessary in order to sustain the
 requested rate.
 
-### report
-```console
-$ vegeta report -h
-Usage of vegeta report:
-  -inputs string
-      Input files (comma separated) (default "stdin")
-  -output string
-      Output file (default "stdout")
-  -reporter string
-      Reporter [text, json, plot, hist[buckets]] (default "text")
-```
+### report command
 
 #### `-inputs`
 Specifies the input files to generate the report of, defaulting to stdin.
@@ -363,17 +334,7 @@ Bucket         #     %       Histogram
 [6ms,   +Inf]  4771  25.93%  ###################
 ```
 
-### `dump`
-```console
-$ vegeta dump -h
-Usage of vegeta dump:
-  -dumper string
-      Dumper [json, csv] (default "json")
-  -inputs string
-      Input files (comma separated) (default "stdin")
-  -output string
-      Output file (default "stdout")
-```
+### `dump` command
 
 #### `-inputs`
 Specifies the input files containing attack results to be dumped. You can specify more than one (comma separated).
--- a/README.md
+++ b/README.md
@@ -48,7 +48,7 @@ attack command:
   -duration duration
       Duration of the test [0 = forever]
   -format string
-      Targets format [legacy] (default "legacy")
+      Targets format [legacy, json] (default "legacy")
   -h2c
       Send HTTP/2 requests without TLS encryption
   -header value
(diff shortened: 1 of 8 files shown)


attack.go
```diff
@@ -111,10 +113,20 @@ func attack(opts *attackOpts) (err error) {
 		src = files[opts.targetsf]
 		hdr = opts.headers.Header
 	)
-	if opts.lazy {
-		tr = vegeta.NewLazyTargeter(src, body, hdr)
-	} else if tr, err = vegeta.NewEagerTargeter(src, body, hdr); err != nil {
-		return err
+
+	switch opts.format {
+	case "json":
```
golangci: string `json` has 3 occurrences, make it a constant

mbrookes1304:
@tsenart I like the approach taken here. I think it's much cleaner creating a new format than trying to find an ugly way to fit this feature into the `legacy` format

README.md
```diff
@@ -172,6 +136,66 @@ The internal concurrency structure's setup has this value as a variable.
...
+Each target is one JSON object in its own line. If present, the body field must be base64 encoded.
+
+```bash
+jq -ncM '{method: "GET", url: "http://goku", body: "Punch!" | @base64, header: {"Content-Type": ["text/plain"]}}' |
+  vegeta attack -format=json -rate=100 | vegeta dump
+```
+
+##### `legacy` format
+
+The ill-defined legacy format almost resembles the plain-text HTTP message format but
```
xla: In here `ill-defined` seems self-deprecating, could be left out.

attack.go
```diff
@@ -25,6 +25,7 @@ func attackCmd() command {
 
 	fs.StringVar(&opts.name, "name", "", "Attack name")
 	fs.StringVar(&opts.targetsf, "targets", "stdin", "Targets file")
+	fs.StringVar(&opts.format, "format", "legacy", "Targets format [legacy, json]")
```
xla: Defaults could be constants.

attack.go
xla: Agree, should be constants.

attack.go
```diff
@@ -111,10 +113,20 @@ func attack(opts *attackOpts) (err error) {
...
-		tr = vegeta.NewLazyTargeter(src, body, hdr)
-	} else if tr, err = vegeta.NewEagerTargeter(src, body, hdr); err != nil {
-		return err
+
+	switch opts.format {
+	case "json":
+		tr = vegeta.NewJSONTargeter(src, body, hdr)
+	case "legacy":
+		fallthrough
+	default:
```
xla: The user should get feedback for wrong inputs, would move this up to the case where it belongs and error for everything else.

README.md
```diff
@@ -172,6 +136,66 @@ The internal concurrency structure's setup has this value as a variable.
...
 responses delay. Use 0 for an infinite attack.
 
+#### `-format`
+Specifies the targets format to decode.
+
+##### `json` format
+
+The JSON format makes integration with programs that produce targets dynamically easier.
+Each target is one JSON object in its own line. If present, the body field must be base64 encoded.
+
```
xla: This section should be expanded with the fileds of the json objects, which are mandatory and which are optional.

lib/targets.go
```diff
@@ -17,10 +18,10 @@ import (
...
 // Target is an HTTP request blueprint.
 type Target struct {
-	Method string
-	URL    string
-	Body   []byte
-	Header http.Header
+	Method string      `json:"method"`
+	URL    string      `json:"url"`
+	Body   []byte      `json:"body"`
+	Header http.Header `json:"header"`
```
xla: Always uncertain which one is canonical `Header` or `Headers`.

attack.go
tsenart: OK, I'll make some constants for these values.

tsenart:
@xla: So, what do you prefer? `text`? I feel like it is too general... The format is "almost" like HTTP. Could we find something along those lines? Would calling it `http` be too much of a lie? What about `httpish`? 😺

README.md
tsenart: OK, agree there's no need to deprecate it. I'll change the wording.

attack.go
tsenart: They could. I'll make this change together with the other constant related changes.

attack.go
tsenart: Yep, this implementation was rather incomplete in view of getting earlier feedback. I'll add proper validation as well as tests.

README.md
tsenart: +1. What do you think of checking in a JSON schema file for this? Is it over-kill?

README.md
```diff
@@ -262,17 +253,7 @@ Specifies the initial number of workers used in the attack. The actual
...
-$ vegeta report -h
-Usage of vegeta report:
-  -inputs string
-      Input files (comma separated) (default "stdin")
-  -output string
-      Output file (default "stdout")
-  -reporter string
-      Reporter [text, json, plot, hist[buckets]] (default "text")
-```
+### report command
```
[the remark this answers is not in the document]
tsenart: Because it was redundant. It's already included in the output of `vegeta -h`.

1u0:
Imo, the drawback of proposed api is that body payloads are not in plain text (base64 encoded).
If you remove base64, but still keep json format as "envelope", then for json body payloads one needs to escape them.
Both things reduce human readability of targets, text searching.

As an alternative, maybe yaml format can be used as an "envelope", this way allowing to put unescaped body payloads.

Or, you may re-consider the idea with HTTP/1.x message format. For example, the approach used in REST client tools (with inline body payloads and `###` as requests separator):
 * [IntelliJ IDEA](https://blog.jetbrains.com/phpstorm/2017/09/editor-based-rest-client/);
 * Visual Studio Code plugin: https://github.com/Huachao/vscode-restclient

This would also make `vegeta` interop with those tools as a bonus.

tsenart:
@1u0 Thanks for bringing that seemingly popular format to our attention! I wasn't aware of it personally. Do you know if it's more formally specified somewhere? What are its origins?

The only issue I can see is that framing scheme doesn't allow for request bodies to contain that same separator in them. That's one issue the Base64 encoding in the JSON format solves. How problematic do you think it could be in practice? I can imagine writing programs to generate Vegeta targets dynamically with request body contents they don't know and hitting this potentially surprising issue. 

Regarding the JSON format, it isn't clear from what you wrote why you find the need to Base64 encode the request bodies a drawback. Could you show us how it hinders you so that we understand your context better?

1u0:
> Regarding the JSON format, it isn't clear from what you wrote why you find the need to Base64 encode the request bodies a drawback.

A potential use case can be that payload was generated (persisted into a file) and passed to `vegeta`. But later on, the payload needs to be manually analysed/checked. With plain text, one may just use common tools (like grep, less) to read and search the payload file.
With your proposal it should be also possible to achieve this, for example by adding `jq` to "unwrap" the payload. It's just a small inconvenience compared with "plain text" approach.

#### About the `.http` file format:

> Do you know if it's more formally specified somewhere? What are its origins?

I'm not aware about any formal specification of it.
My guess that JetBrains came up with this format in their (proprietary) plugin, also inspired by HTTP message format (they introduced it in their blog posts). The plugin for VS Code, looks like, tries to follow the semantics as close as possible to one in IntelliJ IDEA.

> The only issue I can see is that framing scheme doesn't allow for request bodies to contain that same separator `###` in them.

True, but this is limited only at the start of new lines. There is also a potentially conflicting "`< `-rule" to include request body from external file. Both patterns can conflict with real payload.
Imo, it's a design tradeoff.

tsenart:
> A potential use case can be that payload was generated (persisted into a file) and passed to vegeta. But later on, the payload needs to be manually analysed/checked. With plain text, one may just use common tools (like grep, less) to read and search the payload file.
With your proposal it should be also possible to achieve this, for example by adding jq to "unwrap" the payload. It's just a small inconvenience compared with "plain text" approach.

You don't need `jq` to unwrap the payload. Most UNIX environments have a dedicated tool for that. In macOS, that's called `base64`. I understand your nuisance, but I think it's a minor one, and these tools are easy to use and compose in the terminal. Additionally, if you have a program generating targets, any programming language has a base64 module that's easy to use.

> True, but this is limited only at the start of new lines. There is also a potentially conflicting "<-rule" to include request body from external file. Both patterns can conflict with real payload.
Imo, it's a design tradeoff.

Yeah, indeed a tradeoff. I suppose if it's well documented, then it shouldn't be a problem for people. So I'd be on-board with *also* adding this functionality to the HTTP-like format, but I'm wondering if breaking back-wards compatibility in this particular case is worth it. I'm not very sure. What's your take @xla?

README.md
xla: That's a great idea, especially with such small surface,maintaining such a file will be great for documentation and potential integrations.

lib/targets.go
xla: Usually I consult the rfc and it's consistently referring to that part of request/response as header, so all good here. :>

xla:
> @xla: So, what do you prefer? text? I feel like it is too general... The format is "almost" like HTTP. Could we find something along those lines? Would calling it http be too much of a lie? What about httpish?

Ha! `httpy` maybe. :japanese_ogre: I think `http` is fine as long as the caveats are well-documented.

xla:
> Yeah, indeed a tradeoff. I suppose if it's well documented, then it shouldn't be a problem for people. So I'd be on-board with also adding this functionality to the HTTP-like format, but I'm wondering if breaking back-wards compatibility in this particular case is worth it. I'm not very sure. What's your take @xla?

I'd focus on the proposed change for the json format at first and keep the additonal changes to the existing format as potential improvement for the backlog and evaluate in depth of the change is indeed breaking.

The disucssion around encoding of payload bodies seems moot as any sensible human interaction with formats like json needs specific tooling, as much as it is necessary to interact with large amounts of newline delimited plain text. It should not so much inform the design of the implementation as it seems reasonable to assume that the majoirty of interactions with the json definitions come from machine owned pipelines, basically supporting your claim @tsenart that json targets are generated for the most part.

tsenart:
> I'd focus on the proposed change for the json format at first and keep the additonal changes to the existing format as potential improvement for the backlog and evaluate in depth if the change is indeed breaking.

+1

internal/cmd/jsonschema/main.go
```diff
@@ -0,0 +1,61 @@
...
+		"Target": &vegeta.Target{},
+	}
+
+	valid := strings.Join(keys(types), ", ")
+
+	fs := flag.NewFlagSet("jsonschema", flag.ExitOnError)
+	typ := fs.String("type", "", fmt.Sprintf("Vegeta type to generate a JSON schema for [%s]", valid))
+	out := fs.String("output", "stdout", "Output file")
+
+	fs.Parse(os.Args[1:])
```
golangci: Error return value of `fs.Parse` is not checked

tsenart:
@xla: Assigned to you. PTAL again.

attack.go
```diff
@@ -25,6 +26,8 @@ func attackCmd() command {
 
 	fs.StringVar(&opts.name, "name", "", "Attack name")
 	fs.StringVar(&opts.targetsf, "targets", "stdin", "Targets file")
+	fs.StringVar(&opts.format, "format", vegeta.HTTPTargetFormat,
+		fmt.Sprintf("Targets format [%s]", strings.Join(vegeta.TargetFormats, ", ")))
```
xla: I find this breaking odd, it doesn't give consistent guidance how to break long lines and makes it hard to edit. Might be preferrable to use the following rule:

``` go
func myFunc(
	param1,
	param2,
	param3,
)
```

attack.go
```diff
@@ -111,10 +115,23 @@ func attack(opts *attackOpts) (err error) {
...
-	} else if tr, err = vegeta.NewEagerTargeter(src, body, hdr); err != nil {
-		return err
+
+	switch opts.format {
+	case vegeta.JSONTargetFormat:
+		tr = vegeta.NewJSONTargeter(src, body, hdr)
+	case vegeta.HTTPTargetFormat:
+		tr = vegeta.NewHTTPTargeter(src, body, hdr)
+	default:
+		return fmt.Errorf("format %q isn't one of [%s]",
```
xla: Same as above.

lib/targets.go
```diff
@@ -40,17 +43,128 @@ func (t *Target) Request() (*http.Request, error) {
...
 
+// Equal returns true if the target is equal to the other given target.
+func (t *Target) Equal(other *Target) bool {
+	switch {
+	case t == other:
+		return true
+	case t == nil || other == nil:
+		return false
+	default:
+		equal := t.Method == other.Method &&
```
xla: Guess this is for readability, would argue just as hard to read in the if statement itself.

tsenart:
This feature has been looooooong in the making. Since #148, namely. Feels good to make progress again :-)

tsenart:
By feature, I mean in-line response bodies which JSON targets make easy peasy.

omjego:
Hey @tsenart , what's the status of this feature? Is it merged already?

tsenart:
Yes.

PR #308, opened 2018-07-10
feature/override_resolvers: add -resolvers flag to attack command

nathanejohnson:
This adds the ability to override the host operating system resolver
by allowing this to be specified on the command line (comma separated list).

In the event more than one is specified, the resolvers are tried in order in a round robin manner.

#### Background

In some situations it is desirable to override the default host resolver and use a separate host for DNS resolution.  This adds the option to specify the dns server on the command line.

#### Checklist


lib/resolver_test.go
```diff
@@ -0,0 +1,118 @@
...
+	for subtest, tdata := range table {
+		t.Run(subtest, func(t *testing.T) {
+			addrs, err := normalizeResolverAddresses(tdata.input)
+			if tdata.expectError {
+				if err == nil {
+					t.Error("expected error, got none")
+				}
+				return
+
+			} else {
```
golangci: if block ends with a return statement, so drop this else and outdent its block

tsenart:
Hey! Thanks for contribution. Would you please demonstrate with more detail which sort of situation you’re facing that drove this change?

Apart from that, one initial comment:

Would there be a way of avoiding the mutation of the global package level net.DefaultResolver? I’d rather parameterise this in the Attacker only.

lib/resolver.go
```diff
@@ -0,0 +1,82 @@
...
+			return nil, fmt.Errorf("host %s is not an IP address", host)
+		}
+
+		normalAddresses[i] = fmt.Sprintf("%s:%d", host, port)
+	}
+	return normalAddresses, nil
+}
+
+func (r *resolver) dial(ctx context.Context, network, _ string) (net.Conn, error) {
+	return r.dialer.DialContext(ctx, network, r.address())
```
tsenart: A small comment explaining why we’re ignoring the argument would be great

lib/resolver.go
```diff
@@ -0,0 +1,82 @@
...
+	dialer    *net.Dialer
+	idx       uint64
+}
+
+// NewResolver - create a new instance of a dns resolver for plugging
+// into net.DefaultResolver.  Addresses should be a list of
+// ip addresses and optional port numbers, separated by colon.
+// For example: 1.2.3.4:53 and 1.2.3.4 are both valid.  In the absence
+// of a port number, 53 will be used instead.
+func NewResolver(addresses []string) (*net.Resolver, error) {
```
tsenart: This should not be public. It’s not core to Vegeta as a library. Also, isn’t there any library out there that does this exactly?

nathanejohnson:
For our use case, we have several 10GB interfaces on the same machine, and we want to try to control egress.  We have two different "fake" dns servers that always return a single IP, and we use this in combined with the source address option to effectively steer the traffic out one interface or the other.

As for your other point about the net package global, there is really no other way that I'm aware of to influence how net.Dial etc does dns resolution.  That's one reason I made sure that setting that package global is assigned inside the main() and not in the lib package.

lib/resolver.go
nathanejohnson: Not that I'm aware of.  Would you prefer this be moved into main?  Or would you prefer it be moved into a separate subpackage?

lib/resolver.go
nathanejohnson: I will add it to the code and push it pending what you decide on the comment below, but in the mean time, I'm ignoring that value since we're overriding it.  that parameter represents the address of the dns resolver.

nathanejohnson:
@tsenart I have moved the resolver out of the lib package and into a new package called "resolver".  Is that more along the lines of what you wanted?

Regarding the package global, an alternative would be to specify the Resolver in your net.Dialer that is defined in the Attacker here:

https://github.com/tsenart/vegeta/blob/master/lib/attack.go#L52

I suppose I could create an opts func that allows you to set the Resolver portion of the dialer, but one issue is I wouldn't be able to handle errors, such as a malformed host:port specified.  I could just skip problematic hosts I guess?

nathanejohnson:
@tsenart https://github.com/golang/go/issues/12503

One of the workarounds suggested in this issue is very similar to the approach taken here.  They also use the net.DefaultDialer:

https://godoc.org/github.com/benburkert/dns/init

nathanejohnson:
per that same thread, I disabled this functionality on windows since swapping net.DefaultDialer on that platform has no effect (as is documented in the net package):

    On Windows, the resolver always uses C library functions, such as GetAddrInfo and DnsQuery.

tsenart:
> For our use case, we have several 10GB interfaces on the same machine, and we want to try to control egress. We have two different "fake" dns servers that always return a single IP, and we use this in combined with the source address option to effectively steer the traffic out one interface or the other.

I see. As a curiosity, is there anything preventing you from changing your environment's configured DNS resolvers outside Vegeta? Are you using containers (Docker et. all)?

> @tsenart I have moved the resolver out of the lib package and into a new package called "resolver". Is that more along the lines of what you wanted?

This is still public unfortunately. I'd move this code to the main CLI package and make it fully private.

> I suppose I could create an opts func that allows you to set the Resolver portion of the dialer, but one issue is I wouldn't be able to handle errors, such as a malformed host:port specified. I could just skip problematic hosts I guess?

You should handle those errors before setting the resolver in the Attacker. Something like:

```
resolver, err := newResolver(addrs...)
if err != nil {
    return fmt.Errorf("resolver: %v", err)
}

atk := vegeta.NewAttacker(
    vegeta.Resolver(resolver),
)
```

> EDIT - that's not going to work very well because, for instance, H2C() will change the transport, so I can't modify the dialer if someone puts the resolver in the opts chain after H2C(). It really does need to be in the net package global.

`H2C` reuses the original `Dialer` so that shouldn't be a problem. Am I misreading the code?

attack_nonwindows.go
```diff
@@ -0,0 +1,9 @@
+// +build !windows
+
+package main
+
+import "flag"
+
+func systemSpecificFlags(fs *flag.FlagSet, opts *attackOpts) {
+	fs.Var(&opts.resolvers, "resolvers", "Override system dns resolution (comma separated list)")
```
tsenart: "List of addresses (ip:port) to use for DNS resolution. Disables use of local system DNS."

resolver/resolver.go
```diff
@@ -0,0 +1,84 @@
...
+	}
+	return &net.Resolver{
+		PreferGo: true,
+		Dial:     (&resolver{addresses: normalAddresses, dialer: &net.Dialer{}}).dial,
+	}, nil
+}
+
+func normalizeResolverAddresses(addresses []string) ([]string, error) {
+	if len(addresses) == 0 {
+		return nil, errors.New("must specify at least resolver address")
```
tsenart: This check doesn't belong here. It should be in `NewResolver`.

resolver/resolver.go
```diff
@@ -0,0 +1,84 @@
...
+	}
+	normalAddresses := make([]string, len(addresses))
+	for i, addr := range addresses {
```
tsenart: You can re-use https://golang.org/pkg/net/#SplitHostPort for this purpose.

resolver/resolver.go
```diff
@@ -0,0 +1,84 @@
...
+		if ip == nil {
+			return nil, fmt.Errorf("host %s is not an IP address", host)
+		}
+
+		normalAddresses[i] = fmt.Sprintf("%s:%d", host, port)
+	}
+	return normalAddresses, nil
+}
+
+// ignore the third parameter, as this represents the dns server address that
```
tsenart: that **we** are

tsenart: or, "is being ..."

resolver/resolver.go
```diff
@@ -0,0 +1,84 @@
...
+	return r.dialer.DialContext(ctx, network, r.address())
+}
+
+func (r *resolver) address() string {
+	var address string
+	if l := uint64(len(r.addresses)); l > 1 {
+		idx := atomic.AddUint64(&r.idx, 1)
+		address = r.addresses[idx%l]
+	} else {
+		address = r.addresses[0]
```
tsenart: Seems like a premature optimisation to me. I'd rather avoid the special case here.

```go
return r.addrs[atomic.AddUint64(&r.idx, 1)%uint64(len(r.addrs))]
```

resolver/resolver_test.go
```diff
@@ -0,0 +1,118 @@
+package resolver
+
+import (
+	"net"
+	"net/http"
+	"testing"
+)
+
+func TestResolve8888(t *testing.T) {
+	r, err := NewResolver([]string{"8.8.8.8:53"})
```
tsenart: I'd rather not call out Google every time I run the test suite :-) Can we setup a fake DNS server instead? https://godoc.org/github.com/miekg/dns

resolver/resolver_test.go
```diff
@@ -0,0 +1,118 @@
...
+		},
+		"Parse error list": {
+			input: []string{
+				"abcd.com:53",
+			},
+			expectError: true,
+		},
+	}
+	for subtest, tdata := range table {
+		t.Run(subtest, func(t *testing.T) {
```
tsenart: Can these sub tests be ran in `t.Parallel()`?

resolver/resolver_test.go
```diff
@@ -0,0 +1,118 @@
...
+
+	resp.Body.Close()
+}
+
+func TestResolveAddresses(t *testing.T) {
+	table := map[string]struct {
+		input          []string
+		want           []string
+		expectError    bool
+		expectMismatch bool
```
tsenart: This test is testing `normalizeAddresses`, not address resolution as the name implies. With the reuse of the standard library function, I don't see much value in testing this. However, if we were to do it, testing for specific errors and addresses to be returned is the way to go (instead of these booleans).

resolver/resolver_test.go
```diff
@@ -0,0 +1,118 @@
...
+				t.Errorf("unexpected mismatch, input: %#v, want: %#v", addrs, tdata.want)
+			}
+
+		})
+	}
+
+}
+
+func TestResolverOverflow(t *testing.T) {
+	res := &resolver{
```
tsenart: This test also seems unnecessary to me. The code in that method is trivial and common.

resolver/resolver.go
nathanejohnson: Why?

resolver/resolver.go
nathanejohnson: That gives an error unless both host and port are present.  My version allows for just host or combination of host:port

resolver/resolver.go
nathanejohnson: I will add a benchmark

resolver/resolver_test.go
nathanejohnson: I was trying to avoid adding deps unnecessarily.  But if you are fine with that, the thought had already crossed my mind.

resolver/resolver_test.go
nathanejohnson: why?

resolver/resolver_test.go
nathanejohnson: see above

resolver/resolver_test.go
nathanejohnson: I can remove it

resolver/resolver.go
tsenart: Yeah, but you can ignore the error. The host and port are still returned.

resolver/resolver.go
tsenart: Seems like a better place to me. `NewResolver` can't really proceed sanely unless there are addresses passed to it. `normalizeResolverAddresses` however, is a function from `[]string` to `[]string` and hence can simply return an empty slice for an empty input.

resolver/resolver.go
tsenart: I know it'll be faster, but it's not worth it.

resolver/resolver_test.go
tsenart: +1

resolver/resolver_test.go
tsenart: Reduced test run-time.

resolver/resolver.go
nathanejohnson: No, that's not how it works at all.

https://play.golang.org/p/0rTnN5eoAj2

tsenart:
```go
for i, addr := range addrs {
    host, port, _ := net.SplitHostPort(addr)
    if port == "" {
        port = "53"
    }
    normal[i] = net.JoinHostPort(host, port)
}
```

tsenart: Woops, never mind. I was quite convinced that was the case but it isn't.

nathanejohnson: if there is no colon in the hostport string passed to SplitHostPort, host will be blank, port will be blank, error is the only part of the return with non-zero value.

tsenart: Then can we ensure the addr has a port and then reuse `net.SplitHostPort`?

nathanejohnson: sure

tsenart: I meant `net.JoinHostPort`, instead of `fmt.Sprintf`. So, last requests here:

1. Reuse `net.JoinHostPort`
2. Rename `pu16` to `pn` (port number)
3. Rename `normalAddresses` to `normal`

resolver/resolver.go
```diff
@@ -0,0 +1,84 @@
...
+	"errors"
+	"fmt"
+	"net"
+	"strconv"
+	"strings"
+	"sync/atomic"
+)
+
+type resolver struct {
+	addresses []string
```
tsenart: Can we rename this to `addrs`?

resolver/resolver.go
nathanejohnson: Actually it just occurred to me, my naive implementation breaks on parsing ipv6.  So far all of our labs are ipv4

resolver/resolver_test.go
tsenart: OK, cool, then rename the test name to `TestNormalizeAddresses` and structure the test table as:

```go
struct {
    in []string
    out []string
    err error
}
```

With the assertions being something like:

```go

out, err := normalizeAddrs(tc.in)
if got, want := out, tc.out; !reflect.DeepEqual(got, want) {
    t.Errorf("got addrs %v, want %v", got, want)
}

if got, want := fmt.Sprint(err), fmt.Sprint(tc.err); got != want {
    t.Errorf("got err %v, want %v", got, want)
}
```

resolver/resolver.go
tsenart: @nathanejohnson: Ah, well spotted!

tsenart:
Please wait until the end to squash so that everybody can follow along your changes.

resolver/resolver_test.go
```diff
@@ -0,0 +1,171 @@
...
+		err := cmd.Wait()
+		if err != nil {
+			t.Logf("unclean shutdown of dnsmasq: %s", err)
+		}
+		t.Log(stdout.String())
+		t.Log(stderr.String())
+	}()
+	time.Sleep(time.Second)
+
+	defer cmd.Process.Kill()
```
golangci: Error return value of `cmd.Process.Kill` is not checked

nathanejohnson:
@tsenart I made some updates.  Instead of using miekg/dns package as a dns server for use during testing, I decided instead to optionally use the dnsmasq which is available in travis.  To enable that resolver test, the VEGETA_TESTDNSMASQ_ENABLE environment variable must be present.  I updated the travis build to install dnsmasq and set that environment variable.  It's not my favorite solution, but using his dns server package for even a trivial example would have added hundreds of lines to the test, and if you want I can disable running that test within the travis build.

tsenart:
@nathanejohnson: I've used `miekg/dns` before and I find it surprising you found this simple use case to add so much code to our codebase. Have you had a look at introductory articles such as this https://jameshfisher.com/2017/08/04/golang-dns-server.html?

I don't think it's worth it to introduce a run-time dependency like `dnsmasq`. I'm OK with a code dependency like `miekg/dns`.

tsenart:
@nathanejohnson: Do you need any help on bringing this forward?

nathanejohnson:
@tsenart @xla sorry for the radio silence, been covered up with work.  Hopefully I can get back to this on the weekend.

tsenart:
Ping :-)

resolver/resolver_test.go
```diff
@@ -0,0 +1,187 @@
...
+)
+
+func TestResolveMiekg(t *testing.T) {
+
+	dns.HandleFunc(".", func(w dns.ResponseWriter, r *dns.Msg) {
+		m := &dns.Msg{}
+		m.SetReply(r)
+		localIP := net.ParseIP("127.0.0.1")
+		defer func() {
+			w.WriteMsg(m)
```
golangci: Error return value of `w.WriteMsg` is not checked

resolver/resolver_test.go
```diff
@@ -0,0 +1,187 @@
...
+		Addr:         fmt.Sprintf("%s:%s", "127.0.0.1", port),
+		Net:          "udp",
+		UDPSize:      dns.MinMsgSize,
+		ReadTimeout:  2 * time.Second,
+		WriteTimeout: 2 * time.Second,
+		// Unsafe instructs the server to disregard any sanity checks and directly hand the message to
+		// the handler. It will specifically not check if the query has the QR bit not set.
+		Unsafe: false,
+	}
+	go ds.ListenAndServe()
```
golangci: Error return value of `ds.ListenAndServe` is not checked

nathanejohnson:
@tsenart I have updated to use miekg dns

tsenart:
Released: https://github.com/tsenart/vegeta/releases/tag/cli/v11.3.0

PR #318, opened 2018-07-14
dump: Rework command to encode

### LINKED ISSUE #302: Ability to disable response body logging
#### Proposal

Having an additional cli switch that gives the ability to disable response body logging to stdout. 

```
vegeta attack -skip-body 
```

#### Background

We use vegeta for load testing our large scale server solutions with a distributed attacker setup. It is important to simulate real world scenarios therefore the response of the load test requests needs to contain around 2K of data. The down side of this that logs (stdout) produced by the `vegeta` process can potentially become very large (tens of gigabytes of data). As a result these logs become tedious to move around. 

We created a custom build of `vegeta` that allows to simply disable body logging.
https://github.com/Skyscanner/vegeta/commit/4bded67b7e173bd1803e47c6b8a6b280ada50a10

We thought this could be useful for the community as well, therefore, we wonder whether this (or a similar) change could be merged back to the main source.

xla:
In order to give the command more utility and play well with pipeline idea of the vegeta toolchain we decided per discussion[0] to change the semantics of dump. We offer now a general encode command which can transform any supported input encoding to any supported output encoding. We anticipate that in conjunction with advocated auxiliary tools, like jq, a wide range of use-cases can be covered without extending vegeta itself.

[0] https://github.com/tsenart/vegeta/issues/302#issuecomment-403935450<Paste>

Closes #302

#### Checklist

~~- [ ] Changed or added code is covered by appropriate tests.~~


encode.go
```diff
@@ -0,0 +1,114 @@
...
+		to     = fs.String("to", "json", "Output encoding [csv, gob, json]")
+		output = fs.String("output", "", "Output file")
+	)
+
+	fs.Usage = func() {
+		fmt.Println("Usage: vegeta encode [flags] [<file>...]")
+		fs.PrintDefaults()
+	}
+
+	return command{
```
golangci: github.com/tsenart/vegeta.command composite literal uses unkeyed fields

encode.go
```diff
@@ -0,0 +1,114 @@
...
+		fs,
+		func(args []string) error {
+			fs.Parse(args)
```
golangci: Error return value of `fs.Parse` is not checked

encode.go
```diff
@@ -0,0 +1,114 @@
...
+	encodingGob  = "gob"
+	encodingJSON = "json"
+)
+
+type decoderFunc func(io.Reader) vegeta.Decoder
+type encoderFunc func(io.Writer) vegeta.Encoder
+
+func encodeCmd() command {
+	var (
+		fs     = flag.NewFlagSet("encode", flag.ExitOnError)
```
tsenart: This should be `flag.NewFlagSet("vegeta encode", ...)`

encode.go
```diff
@@ -0,0 +1,114 @@
...
+		from   = fs.String("from", "gob", "Input decoding [csv, gob, json]")
+		to     = fs.String("to", "json", "Output encoding [csv, gob, json]")
+		output = fs.String("output", "", "Output file")
```
tsenart: The default output should be `stdout`

encode.go
```diff
@@ -0,0 +1,114 @@
...
+	)
+
+	fs.Usage = func() {
+		fmt.Println("Usage: vegeta encode [flags] [<file>...]")
+		fs.PrintDefaults()
```
tsenart: What do you think of including a small intro to the command. i.e. https://github.com/tsenart/vegeta/blob/0aa7bbf15c9faca2d684872851b1b48bb0d00c07/plot.go#L14

encode.go
tsenart: Ugh, OK, let's comply. Other places are in transgression too.

encode.go
tsenart: Since we introduced @golangci, let's make it happy: `_ = fs.Parse()`

encode.go
```diff
@@ -0,0 +1,114 @@
...
+	encodingJSON = "json"
+)
+
+type decoderFunc func(io.Reader) vegeta.Decoder
+type encoderFunc func(io.Writer) vegeta.Encoder
+
+func encodeCmd() command {
+	var (
+		fs     = flag.NewFlagSet("encode", flag.ExitOnError)
+		from   = fs.String("from", "gob", "Input decoding [csv, gob, json]")
```
tsenart: Can we use the constants above to generate these flag usage strings?

encode.go
```diff
@@ -0,0 +1,114 @@
...
+
+func encode(from, to, output string, inputs ...string) error {
+	var decFn decoderFunc
+
+	switch from {
+	case encodingCSV:
+		decFn = vegeta.NewCSVDecoder
+	case encodingJSON:
+		decFn = vegeta.NewJSONDecoder
+	default:
```
tsenart: Do we need to rely on the default case to play nice with the pipes as per your comment? The default "gob" could come from the flags outside this function.

encode.go
```diff
@@ -0,0 +1,114 @@
...
+		// Gob is our default decode format to play nicely in pipes with attack.
+		decFn = vegeta.NewDecoder
+	}
+
+	decs := []vegeta.Decoder{}
+
+	if len(inputs) > 0 {
```
tsenart: We have some simpler code to handle this already. i.e. https://github.com/tsenart/vegeta/blob/master/report.go#L33

Because the `file` function returns `os.Stdin` for the string `stdin`. This block of code can become:

```
	srcs := make([]vegeta.Decoder, len(files))
	for i, f := range inputs {
		in, err := file(f, false)
		if err != nil {
			return err
		}
		defer in.Close()
		srcs[i] = vegeta.NewDecoder(in)
	}
	dec := vegeta.NewRoundRobinDecoder(srcs...)
```

encode.go
```diff
@@ -0,0 +1,114 @@
...
+			}
+			defer f.Close()
+
+			decs = append(decs, decFn(f))
+		}
+	} else {
+		decs = append(decs, decFn(os.Stdin))
+	}
+
+	in := vegeta.NewRoundRobinDecoder(decs...)
```
tsenart: I am wondering what makes most sense to do. Shall we read results in a round-robin fashion and re-encode each out as they come or should we instead read all of them, sort them and then write them out. Perhaps this should be a flag.

encode.go
```diff
@@ -0,0 +1,114 @@
...
+		encFn = vegeta.NewEncoder
+	default:
+		// JSON is our default encoding format to play nicely in pipes with auxiliary
+		// tools like jq.
+		encFn = vegeta.NewJSONEncoder
+	}
+
+	var out vegeta.Encoder
+
+	if output != "" {
```
tsenart: `output` should never be empty. The default `stdout` ought to be set in the flags. Then you can reuse the `file` function as in the report command.

tsenart: Additionally, if you open the output first **and then** validate the `to` value, you don't need `encoderFunc`.

```go

var enc vegeta.Encoder
switch to {
case encodingCSV:
    enc = vegeta.NewCSVEncoder(out)
case encodingJSON:
    enc = vegeta.NewJSONEncoder(out)
case encodingGob:
    enc = vegeta.NewEncoder(out)
}
```

encode.go
```diff
@@ -0,0 +1,114 @@
...
+			return err
+		}
+		defer f.Close()
+
+		out = encFn(f)
+	} else {
+		out = encFn(os.Stdout)
+	}
+
+	for {
```
tsenart: We need to handle signals gracefully as in the report command.

encode.go
xla: Why? `vegeta` is already the main namespace of the binary.

encode.go
xla: This doesn't add up really. We expect `output` to be a filename, if we expect `stdout` as ivalue it is inconsistent. They don't have the same meaning or refer to the same thing. Arguably we shoul document the default case better.

encode.go
xla: To your first comment see https://github.com/tsenart/vegeta/pull/318#discussion_r202521973

To the second comment I usually try to check all parameters before acquiring system resources (fds, conns, etc.).

encode.go
xla: Will extend!

encode.go
xla: What is the upside of the sorting? And sorting per input, over all inputs, by what criteria?

encode.go
tsenart: > We expect output to be a filename.

Sure, and if you so wished you should be able to pass `/dev/stdout` to it and it'd just work. `stdout` is in this case a symbolic name that resolves to that path in practice.

tsenart: Plus, it documents the behaviour, from the user perspective, it makes it explicit in the output of `vegeta encode -h` that the output will be to stdout, instead of that being implicit.

encode.go
tsenart: Yeah, `file` is slightly conflated. We can take out the file creation part from it and use `os.Create` directly where needed, thus scoping the symbolic resolution to opening files only, not creating (since you don't create `stdout` or `stdin`).

```go
func open(file string) (*os.File, error) {
    switch file {
    case "stdout":
        return os.Stdout, nil
    case "stdin":
        return os.Stdin, nil
    default:
        return os.Open(file)
}
```

encode.go
tsenart: I think we can delegate this to external tools, indeed.

```
cat results.bin | vegeta encode | jq -s 'sort_by(.timestamp)'
```

encode.go
tsenart: > To the second comment I usually try to check all parameters before acquiring system resources (fds, conns, etc.).

What does it gain us here?

encode.go
```diff
@@ -0,0 +1,109 @@
...
+	to := fs.String("to", encodingJSON, "Output encoding "+encs)
+	output := fs.String("output", "stdout", "Output file")
+
+	fs.Usage = func() {
+		fmt.Println("Usage: vegeta encode [flags] [<file>...]")
+		fs.PrintDefaults()
+	}
+
+	return command{fs, func(args []string) error {
+		fs.Parse(args)
```
golangci: Error return value of `fs.Parse` is not checked

encode.go
```diff
@@ -0,0 +1,139 @@
...
+	fs := flag.NewFlagSet("vegeta encode", flag.ExitOnError)
+	to := fs.String("to", encodingJSON, "Output encoding "+encs)
+	output := fs.String("output", "stdout", "Output file")
+
+	fs.Usage = func() {
+		fmt.Fprintln(os.Stderr, encodeUsage)
+	}
+
+	return command{fs, func(args []string) error {
+		fs.Parse(args)
```
golangci: Error return value of `fs.Parse` is not checked

README.md
```diff
@@ -99,13 +99,13 @@ report command:
   -reporter string
       Reporter [text, json, plot, hist[buckets]] (default "text")
 
-dump command:
-  -dumper string
-      Dumper [json, csv] (default "json")
-  -inputs string
-      Input files (comma separated) (default "stdin")
+encode command:
```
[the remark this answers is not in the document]
tsenart: I will implement this in a follow up PR.

PR #326, opened 2018-08-24
attack: Support response body read limits

### LINKED ISSUE #325: Add library option to disable saving response bodies
#### Proposal

Add an option for `Attacker` that does not save the response body into a buffer, but copies it directly to `ioutil.Discard`. I've implemented this in my fork: https://github.com/bluekeyes/vegeta/tree/bluekeyes/discard-body

#### Background

This is similar to #302, but slightly different. I'm using Vegeta as a library to test a download server, with file sizes ranging from 1kb to several GB. While the latest version of the CLI allows me to filter out the body from result files (or I could set the body to `nil` when using the library and processing results), reading large responses quickly uses up available memory during the test. 

For the largest downloads in my test (0.25GB - 4GB), my test containers were running out of memory unpredictably based on the download rate (I tested with up to 8GB of available memory.) Since I don't use the response body for anything, discarding it keeps the memory usage low and allows the large tests to run with the same system resources as tests for smaller files.

#### Workarounds

Running tests on hosts with more memory or distributing the load across more machines can solve this problem, but is wasteful and may be hard to setup.

tsenart:
This change set introduces a `-max-body` flag in the attack command
as well as the corresponding `vegeta.MaxBody` functional option to
use with `vegeta.NewAttacker`.

Fixes #325

### DIFF
--- a/lib/attack_test.go
+++ b/lib/attack_test.go
@@ -236,3 +236,33 @@ func TestProxyOption(t *testing.T) {
 		t.Errorf("got body: %q, want: %q", got, want)
 	}
 }
+
+func TestMaxBody(t *testing.T) {
+	t.Parallel()
+
+	body := []byte("VEGETA")
+	server := httptest.NewServer(
+		http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			w.Write(body)
+		}),
+	)
+	defer server.Close()
+
+	for i := DefaultMaxBody; i < int64(len(body)); i++ {
+		maxBody := i
+		t.Run(fmt.Sprint(maxBody), func(t *testing.T) {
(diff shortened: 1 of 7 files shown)


lib/attack_test.go
```diff
@@ -236,3 +236,33 @@ func TestProxyOption(t *testing.T) {
...
 	}
 }
+
+func TestMaxBody(t *testing.T) {
+	t.Parallel()
+
+	body := []byte("VEGETA")
+	server := httptest.NewServer(
+		http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			w.Write(body)
```
golangci: Error return value of `w.Write` is not checked

Issue #305, opened 2018-07-05
Library overhaul


tsenart:
Vegeta's library usability is OK, but could be substantially improved with better patterns, feedback from users and other lessons learned over the years. This issue tracks ideas for improvement and discussion around them.

### Ideas

- Explore other concurrency models of `Targeters` (i.e. channels instead of locks)
- Get rid of `vegeta.Attacker` which is mostly wrapping an `http.Client`.
- Rename `Result` to `Hit`.
- Abstract underlying HTTP client used with the `Hitter` interface (see below).
- Make `vegeta.Attack` a struct and make it composable with other plans for easy orchestration.
- Support other time units than per-second in rates.

```go
package lib_test

import (
	"bytes"
	"context"
	"errors"
	"fmt"
	"io/ioutil"
	"net/http"
	"net/http/httptest"
	"sync"
	"sync/atomic"
	"time"
)

// ExampleSingleAttack exemplifies the usage of Vegeta with a single attack.
func ExampleSingleAttack() {
	srv := func(counter uint64) *httptest.Server {
		return httptest.NewServer(http.HandlerFunc(
			func(w http.ResponseWriter, r *http.Request) {
				fmt.Fprintln(w, atomic.AddUint64(&counter, 1))
			},
		))
	}(0)

	attack := Attack{
		Name:     "1000 req/s",
		Duration: 10 * time.Second,
		Rate:     Rate{Freq: 1000, Per: time.Second},
		Targeter: NewStaticTargeter(Target{Method: "GET", URL: srv.URL}),
		Hitter:   NewHitter(http.DefaultClient),
		Hits:     make(chan *Hit),
	}

	go Attacker{Workers: 10}.Do(context.Background(), attack)

	for hit := range attack.Hits {
		fmt.Printf("Hit: %#v\n", hit)
	}
}

// ExampleStagedAttack exemplifies the usage of Vegeta with a staged attack,
// where request rate ramps up between each attack.
func ExampleStagedAttack() {
	srv := func(counter uint64) *httptest.Server {
		return httptest.NewServer(http.HandlerFunc(
			func(w http.ResponseWriter, r *http.Request) {
				fmt.Fprintln(w, atomic.AddUint64(&counter, 1))
			},
		))
	}(0)

	var attacks []Attack
	for _, params := range []struct {
		QPS int
		For time.Duration
	}{
		{QPS: 10, For: 5 * time.Second},
		{QPS: 100, For: 30 * time.Second},
[...]

Issue #263, opened 2017-09-26
Expose http.Client


julianvmodesto:
I'd like to set up an http.Client that dials a unix socket. In the Vegeta library, could we expose the http.Client to allow arbitrary options?

tsenart:
How would the proposed API in the examples in https://github.com/tsenart/vegeta/issues/305 work for you?

tsenart:
Fixed by #329

PR #329, opened 2018-08-27
Feature (library): BYO client and dialer

sjauld:
#### Background

Temp solution for #263 whilst #305 is under development

#### Checklist

- [/] Git commit messages conform to [community standards](http://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html).
- [/] Each Git commit represents meaningful milestones or atomic units of work.
- [/] Changed or added code is covered by appropriate tests.

### DIFF
--- a/lib/attack.go
+++ b/lib/attack.go
@@ -204,14 +204,9 @@ func MaxBody(n int64) func(*Attacker) {
 	return func(a *Attacker) { a.maxBody = n }
 }
 
-// Client returns a functional option that allows you to BYO http.Client
-func Client(c http.Client) func(*Attacker) {
-	return func(a *Attacker) { a.client = c }
-}
-
-// Dialer returns a functional option that allows you to BYO *net.Dialer
-func Dialer(d *net.Dialer) func(*Attacker) {
-	return func(a *Attacker) { a.dialer = d }
+// Client returns a functional option that allows you to bring your own http.Client
+func Client(c *http.Client) func(*Attacker) {
+	return func(a *Attacker) { a.client = *c }
 }
 
 // A Rate of hits during an Attack.
--- a/lib/attack.go
+++ b/lib/attack.go
@@ -204,6 +204,16 @@ func MaxBody(n int64) func(*Attacker) {
 	return func(a *Attacker) { a.maxBody = n }
 }
 
+// Client returns a functional option that allows you to BYO http.Client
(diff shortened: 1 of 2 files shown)


lib/attack.go
```diff
@@ -204,6 +204,16 @@ func MaxBody(n int64) func(*Attacker) {
 	return func(a *Attacker) { a.maxBody = n }
 }
 
+// Client returns a functional option that allows you to BYO http.Client
```
tsenart: BYO is unnecessarily abbreviated and may lead to some users needing to lookup the term. Let's err on the side of explicitness here.

lib/attack.go
```diff
@@ -204,6 +204,16 @@ func MaxBody(n int64) func(*Attacker) {
...
+func Client(c http.Client) func(*Attacker) {
+	return func(a *Attacker) { a.client = c }
+}
+
+// Dialer returns a functional option that allows you to BYO *net.Dialer
+func Dialer(d *net.Dialer) func(*Attacker) {
```
tsenart: Given we already allow setting the `http.Client` above, do we need this option? You can set the dialer in the client directly.

lib/attack.go
```diff
@@ -204,6 +204,16 @@ func MaxBody(n int64) func(*Attacker) {
 	return func(a *Attacker) { a.maxBody = n }
 }
 
+// Client returns a functional option that allows you to BYO http.Client
+func Client(c http.Client) func(*Attacker) {
```
tsenart: I think we should pass a pointer here unless you have a specific reason not to.

lib/attack_test.go
```diff
@@ -266,3 +267,73 @@ func TestMaxBody(t *testing.T) {
...
+	tr := NewStaticTargeter(Target{Method: "GET", URL: server.URL})
+	rate := Rate{Freq: 100, Per: time.Second}
+
+	atk := NewAttacker(Client(client))
+	var hits uint64
+	for range atk.Attack(tr, rate, 1*time.Second, "") {
+		hits++
+	}
+
+	if got, want := hits, uint64(rate.Freq); got != want {
```
tsenart: This doesn't actually test that a specific client was used. Perhaps an easy way to do so would be to set a timeout of zero and check for the particular error is returned.

sjauld:
Changes implemented!

tsenart:
Nice, thanks!

PR #341, opened 2018-10-07
attack: Read response bodies in full

### LINKED ISSUE #339: attack: Read full response bodies
https://github.com/tsenart/vegeta/pull/326 introduce a way to limit the number of bytes read from HTTP responses but, as pointed out by @smartyjohn in https://github.com/tsenart/vegeta/issues/302#issuecomment-427054465, it didn't solve the issue of reading all the response bodies, in order to observe the full network latency timing, but discard them at the end.

Marshalling out the responses to only filter the bodies isn't always an option in high scale load tests. So we ought to provide an option to do it at the source.

tsenart:
PR #326 introduced a way to limit the number of bytes read from HTTP responses.
However, as pointed out by @smartyjohn in #302 (comment), it didn't solve the need
of reading all the response bodies (in order to observe the full network latency timing)
but discard them at the end.

This was an oversight and unintentional, so this change set changes the semantics
of both the `-max-body` flag in the `attack` command and the correspondent
`vegeta.MaxBody` `Attacker` option.

Instead of limiting the number of bytes **read** from the response bodies,
it limits the number of bytes **captured** from response bodies while
always reading the full body from the underlying connection.

Fixes #339

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,7 +75,7 @@ attack command:
   -lazy
     	Read targets lazily
   -max-body value
-    	Maximum number of bytes to be read from response bodies. [-1 = no limit] (default -1)
+    	Maximum number of bytes to capture from response bodies. [-1 = no limit] (default -1)
   -name string
     	Attack name
   -output string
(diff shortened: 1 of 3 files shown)


tsenart:
PTAL @smartyjohn, @tamasflamich

tamasflamich:
Hi @tsenar -  Looks good, thank you for reaching out.

tsenart:
Released: https://github.com/tsenart/vegeta/releases/tag/cli/v12.0.0

PR #346, opened 2018-10-23
report: Implement --every flag for live updates

### LINKED ISSUE #345: Support reporting while piping
#### Proposal

Add a flag to `vegeta report` that displays some information while binary data is piped to its STDIN.

#### Background

My current `vegeta` calls look like this:
```
vegeta attack -rate 1 -header 'Authorization: Bearer ...' -targets ~/bouillotte -insecure -duration 20m |pv -rt |vegeta report
```
Where `pv` acts as a time & volume indicator.

I'd rather not wait for the whole attack to finish before having some information which look like it could be parsed & displayed as the attack is running. (But then I may be wrong!)

#### Workarounds

The `pv` trick works fine but doesn't show anything that `vegeta report` shows: only `0:06:13 [34.6KiB/s]`

There's also https://github.com/tsenart/vegeta/issues/299 & the two are not incompatible: `vegeta attack` could detect if it is being piped.

tsenart:
This commit introduces the `--every` flag in the report command which
enables reports to be written out at specified interval. This allows
inspection of results as attacks are happening, instead of after they
finish.

Fixes #345

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ sudo: false
 go:
   - tip
 install:
-  - go get -v github.com/golang/lint/golint
+  - go get -v golang.org/x/lint/golint
   - go get -d -t -v ./...
   - go build -v ./...
 script:
--- a/lib/reporters.go
+++ b/lib/reporters.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"sort"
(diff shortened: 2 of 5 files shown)


tsenart:
>  the flag naming is not very telling as it stands, but have no good idea how to improve that.

It should read like english: vegeta *report* **--every** 100ms

Issue #303, opened 2018-07-05
Capture HTTP response header in vegeta.Result


tsenart:
#### Proposal

Extend `vegeta.Result` with a `Header` field that captures the response header of each performed HTTP request (hit).

#### Background

In addition to capturing response bodies, it can be useful to capture response headers for debugging purposes, dynamic target generation based on live attack results, and any other custom integrations via `vegeta dump`.

##### Related issues

- https://github.com/tsenart/vegeta/issues/247
- https://github.com/tsenart/vegeta/issues/267

#### Workarounds

No.

PR #350, opened 2018-10-26
Header information within result and testing

midnightrun:
#### Background

As mentioned in the issue https://github.com/tsenart/vegeta/issues/303, I added the header information after each attack.

#### Checklist

### DIFF
--- a/lib/attack_test.go
+++ b/lib/attack_test.go
@@ -208,8 +208,8 @@ func TestResponseBodyCapture(t *testing.T) {
 func TestResponseHeaderCapture(t *testing.T) {
 	t.Parallel()
 
-	key := "X-test"
-	value := "vegeta"
+	const key = "X-test"
+	const value = "vegeta"
 
 	server := httptest.NewServer(
 		http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
--- a/lib/attack_test.go
+++ b/lib/attack_test.go
@@ -205,6 +205,25 @@ func TestResponseBodyCapture(t *testing.T) {
 		t.Fatalf("got: %v, want: %v", got, want)
 	}
 }
+func TestResponseHeaderCapture(t *testing.T) {
+	t.Parallel()
+
+	key := "X-test"
+	value := "vegeta"
+
+	server := httptest.NewServer(
+		http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			w.Header().Set(key, value)
+		}),
+	)
+	defer server.Close()
+	atk := NewAttacker()
+	tr := NewStaticTargeter(Target{Method: "GET", URL: server.URL})
+	res := atk.hit(tr, "")
+	if v := res.Header.Get(key); v != value {
+		t.Fatalf("got: %v, want: %v", v, value)
+	}
+}
 
 func TestProxyOption(t *testing.T) {
 	t.Parallel()
--- a/lib/attack.go
+++ b/lib/attack.go
@@ -323,6 +323,8 @@ func (a *Attacker) hit(tr Targeter, name string) *Result {
 		return &res
 	}
 
+	res.Header = r.Header
+
 	res.Latency = time.Since(res.Timestamp)
 	res.BytesIn = uint64(len(res.Body))
 
--- a/lib/results.go
+++ b/lib/results.go
@@ -6,10 +6,15 @@ import (
 	"encoding/base64"
 	"encoding/csv"
 	"encoding/gob"
+	"fmt"
 	"io"
+	"log"
 	"net/http"
+	"net/textproto"
+	"reflect"
 	"sort"
 	"strconv"
+	"strings"
 	"time"
 
 	"github.com/mailru/easyjson/jlexer"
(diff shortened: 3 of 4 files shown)


lib/attack_test.go
```diff
@@ -205,6 +205,25 @@ func TestResponseBodyCapture(t *testing.T) {
 		t.Fatalf("got: %v, want: %v", got, want)
 	}
 }
+func TestResponseHeaderCapture(t *testing.T) {
+	t.Parallel()
+
+	key := "X-test"
```
tsenart: Can we make these constants?

midnightrun:
Hi,
I will fix it on Sunday :)

midnightrun:
Hi Tomás,
I finally found the time to play with my PR. My proposal would be that I concatenated the headers into a string and encode it into the CSV. It would be looking like this:

`Content-Length: 14888 Expires: Mon, 31 Mar 2014 06:45:15 GMT`

The interesting point is that we are dealing here with an two dimensional array and encoding it to a string would be the easiest way, I could think of. Encoding would be the same way back.

Any thoughts?

tsenart:
How would that look in practice, with a bunch of records?

midnightrun:
Hi Tomás,
sure.

```go
package main

import (
	"bufio"
	"fmt"
	"log"
	"net/http"
	"net/textproto"
	"strings"
)

func main() {
	res, err := http.Get("http://www.google.de")

	if err != nil {
		log.Fatal(err)
	}

	headerToString := headerToString(res.Header)
	stringToHeader := stringToHeader(headerToString)

	log.Println(headerToString)
	log.Println(stringToHeader)
}

func stringToHeader(headers string) http.Header {
	reader := bufio.NewReader(strings.NewReader(headers + "\r\n"))
	tp := textproto.NewReader(reader)

	mimeHeader, err := tp.ReadMIMEHeader()
	if err != nil {
		log.Fatal(err)
	}

	header := http.Header(mimeHeader)

	return header
}

func headerToString(header http.Header) string {
	headersCopy := []string{}

	for name, values := range header {

		header := fmt.Sprint(name, ":")
		for _, value := range values {
			header = fmt.Sprint(header, value)
		}

		headersCopy = append(headersCopy, header)
	}

	result := strings.Join(headersCopy, "\r\n")
	result = result + "\r\n"

	return result
}
```

Output

```bash
2018/11/01 22:39:04 P3p:CP="This is not a P3P policy! See g.co/p3phelp for more info."
Server:gws
Set-Cookie:1P_JAR=2018-11-01-21; expires=Sat, 01-Dec-2018 21:39:04 GMT; path=/; domain=.google.deNID=144=m15IdlakyN1-16swT4ePT-_V87xNWcTC4kqFSw2kHDywfxigMSvlsrtdxbCRZ09vbgFZwJPoGedsZzoOf2Xr0Xg98ll2txZQ5XpEe3q00B4LM_9E8z0o9bZu2mwmwTqonlHold59cdV9oORbp3ACpttE_JusgTyXiExD9_7ydxE; expires=Fri, 03-May-2019 21:39:04 GMT; path=/; domain=.google.de; HttpOnly
Content-Type:text/html; charset=ISO-8859-1
Expires:-1
Cache-Control:private, max-age=0
X-Xss-Protection:1; mode=block
X-Frame-Options:SAMEORIGIN
Date:Thu, 01 Nov 2018 21:39:04 GMT

2018/11/01 22:39:04 map[P3p:[CP="This is not a P3P policy! See g.co/p3phelp for more info."] Cache-Control:[private, max-age=0] X-Xss-Protection:[1; mode=block] Date:[Thu, 01 Nov 2018 21:39:04 GMT] Server:[gws] Set-Cookie:[1P_JAR=2018-11-01-21; expires=Sat, 01-Dec-2018 21:39:04 GMT; path=/; domain=.google.deNID=144=m15IdlakyN1-16swT4ePT-_V87xNWcTC4kqFSw2kHDywfxigMSvlsrtdxbCRZ09vbgFZwJPoGedsZzoOf2Xr0Xg98ll2txZQ5XpEe3q00B4LM_9E8z0o9bZu2mwmwTqonlHold59cdV9oORbp3ACpttE_JusgTyXiExD9_7ydxE; expires=Fri, 03-May-2019 21:39:04 GMT; path=/; domain=.google.de; HttpOnly] Content-Type:[text/html; charset=ISO-8859-1] Expires:[-1] X-Frame-Options:[SAMEORIGIN]]
```

Like this we had a simply string we could encode into CSV and back to `http.Header`

Sorry, for the extra `\r\n` in the `headerToString` function but `ReadMimeHeader` seems to need an extra blank line to work.

Chéers!

tsenart:
Doesn't this break the encoding and decoding of `vegeta encode`? I'd like to see a test that shows how the encoding and decoding round-trips correctly.

midnightrun:
Okay,
it seems that you were totally right. Now everything is breaking after my change. I think I need some advice here :)

I adjusted the test to also accept the Header property. In the En- and Decoder I added it to, still running into errors :/

midnightrun:
Hi Tomás,
I found out that the problem is the JSON encoding. I am out it, don't worry, it will be fixed :)

tsenart:
Ping.

PR #379, opened 2019-03-05
docs(cpus): correct default of CPUs usage

mstruebing:
#### Background

In the Readme is stated that the number of CPUs used defaults to 8 which isn't correct, see: https://github.com/tsenart/vegeta/blob/0cabfc957a3db953c27ffaccff3ac2327a0dbcde/main.go#L24

BTW: is the commit message okay or how should I name it?

#### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,7 +51,7 @@ Usage: vegeta [global flags] <command> [command flags]
 
 global flags:
   -cpus int
-    	Number of CPUs to use (default 8)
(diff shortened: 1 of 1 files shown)


tsenart:
This is just the output of running the command in my machine.

mstruebing:
But on another machine it would be a different number, so I've updated the readme correctly.

PR #380, opened 2019-03-07
Create Rater interface to allow for attack rates to vary on the fly.

fluffle:
Hi there!

#### Background

I wanted to be able to vary attack rates over time, for Reasons. The easiest way of doing so seems to be to create an interface that can provide the information Attack needs, namely the total number of hits in the provided Duration, and the interval to sleep before the next hit. The change is primarily structured so that code using the vegeta lib will continue to function as before. I'm not wedded to the "Rater" interface name.

If you'd like to know more about *why* I want this, please let me know.

#### Checklist

- [✗] Git commit messages conform to [community standards](http://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html).

Hrrrm, I might have to amend and re-push this branch to meet this requirement, I'll do that after I'm finished typing here.

- [✓] Each Git commit represents meaningful milestones or atomic units of work.

- [✓] Changed or added code is covered by appropriate tests.

All tests of attack behaviour still pass. If you'd like me to add specific unit tests for Hits / Interval, please let me know.

### DIFF
--- a/lib/attack.go
+++ b/lib/attack.go
@@ -223,58 +223,6 @@ func Client(c *http.Client) func(*Attacker) {
 	return func(a *Attacker) { a.client = *c }
 }
 
-// Rater defines the rate of hits during an Attack.
-type Rater interface {
-	// Interval returns the time the attacker needs to sleep
-	// before the next hit is sent to the target.
-	Interval(began time.Time, count uint64) time.Duration
-	// Hits takes the desired attack duration and returns the
-	// number of hits sent in that duration while attacking.
-	Hits(attackDuration time.Duration) uint64
-	// IsZero returns true if the rater is zero-valued.
-	IsZero() bool
-}
-
-// Rate sends a constant rate of hits to the target.
-type Rate struct {
-	Freq int           // Frequency (number of occurrences) per ...
-	Per  time.Duration // Time unit, usually 1s
-}
-
-// IsZero returns true if either Freq or Per are zero valued.
-func (r Rate) IsZero() bool {
-	return r.Freq == 0 || r.Per == 0
-}
-
-// Interval implements part of the Rater interface. It calculates the time
-// between each hit from r.Per / r.Freq and multiplies this by the count of
-// hits already elapsed to determine the absolute time when the next hit should
-// occur. It returns the Duration until that time.
-func (r Rate) Interval(began time.Time, count uint64) time.Duration {
-	return r.interval(began, time.Now(), count)
-}
-
-func (r Rate) interval(began, now time.Time, count uint64) time.Duration {
-	delta := time.Duration(count * uint64(r.Per.Nanoseconds()/int64(r.Freq)))
-	return began.Add(delta).Sub(now)
-}
-
-// Hits implements part of the Rater interface. It returns the number of hits
-// the attacker is expected to send when applying this Rate over the provided
-// Duration.
-func (r Rate) Hits(du time.Duration) uint64 {
-	if du == 0 || r.IsZero() {
-		return 0
-	}
-	return uint64(du) / (uint64(r.Per.Nanoseconds() / int64(r.Freq)))
-}
-
-// String returns a pretty-printed description of the rate, e.g.:
-//   Rate{1 hits/1s} for Rate{Freq:1, Per: time.Second}
-func (r Rate) String() string {
-	return fmt.Sprintf("Rate{%d hits/%s}", r.Freq, r.Per)
-}
-
 // Attack reads its Targets from the passed Targeter and attacks them at
 // the rate specified for the given duration. When the duration is zero the attack
 // runs until Stop is called. Results are sent to the returned channel as soon
--- a/lib/attack.go
+++ b/lib/attack.go
@@ -72,8 +72,8 @@ func NewAttacker(opts ...func(*Attacker)) *Attacker {
 
 	a.client = http.Client{
 		Transport: &http.Transport{
-			Proxy:                 http.ProxyFromEnvironment,
-			Dial:                  a.dialer.Dial,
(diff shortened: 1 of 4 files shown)


lib/attack.go
```diff
@@ -71,8 +71,8 @@ func NewAttacker(opts ...func(*Attacker)) *Attacker {
 
 	a.client = http.Client{
 		Transport: &http.Transport{
-			Proxy:                 http.ProxyFromEnvironment,
-			Dial:                  a.dialer.Dial,
+			Proxy: http.ProxyFromEnvironment,
+			Dial:  a.dialer.Dial,
```
LasTshaMAN: Looks like `go fmt` should be used here ?

lib/attack.go
```diff
@@ -222,11 +234,26 @@ func (r Rate) IsZero() bool {
 	return r.Freq == 0 || r.Per == 0
 }
 
+// Interval returns the time between now and the next hit.
```
LasTshaMAN: Don't you think that docs for `Rater.Interval` and `Rate.Interval` should be that same ? 

I mean, reading both `Interval returns the time the attacker needs to sleep before the next hit is sent to the target.` and `Interval returns the time between now and the next hit.` it is not quite clear that they are equivalent (and if they are, it would be better to just put it in one place - just document interface)

LasTshaMAN:
`If you'd like me to add specific unit tests for Hits / Interval, please let me know.`

Well, you should test your code by default (otherwise how would you know that it works ?). Also, tests will serve as a useful documentation ...

lib/attack.go
fluffle: My editor runs `goimports` on save. I imagine this file hasn't been fmt'd since the go1.11 formatting changes.

lib/attack.go
fluffle: They are approximately the same because Rate implements Rater. I can nuke this comment if you want, but then `go vet` will then complain about an exported function that is undocumented, and your documentation will be incomplete.

Other options:

  - Make the comments *exactly* the same, rather than just mostly the same.
  - Change the comment to read something like:

        Interval implements part of the Rater interface. It calculates the time between each hit
        from `r.Per / r.Freq` and multiplies this by the count of hits already elapsed to determine
        the absolute time when the next hit should occur. It returns the Duration until that time.

LMK what you'd prefer I do?

fluffle:
> Well, you should test your code by default (otherwise how would you know that it works ?). Also, tests will serve as a useful documentation ...

It's not my code, really ;-)

All I've done is move a bunch of `Rate`-specific code out of `attack`. The tests that test the behaviour of `attack` still exist and still pass, even though some of the code is now in separate methods.

But I'll take your comment as an indication that separate unit tests would be appreciated, I'll get on that right now.

fluffle:
PTAL at the new diff.

lib/attack.go
LasTshaMAN: Yeah, you are right ;)

LasTshaMAN:
> > Well, you should test your code by default (otherwise how would you know that it works ?). Also, tests will serve as a useful documentation ...
> 
> It's not my code, really ;-)
> 
> All I've done is move a bunch of `Rate`-specific code out of `attack`. The tests that test the behaviour of `attack` still exist and still pass, even though some of the code is now in separate methods.
> 
> But I'll take your comment as an indication that separate unit tests would be appreciated, I'll get on that right now.

That looks much better ;)

lib/attack.go
```diff
@@ -239,12 +280,10 @@ func (a *Attacker) Attack(tr Targeter, r Rate, du time.Duration, name string) <-
 		defer close(results)
 		defer workers.Wait()
 		defer close(ticks)
-		interval := uint64(r.Per.Nanoseconds() / int64(r.Freq))
-		hits := uint64(du) / interval
+		hits := r.Hits(du)
```
tsenart: If the rate can be variable, we cannot know the total number of hits in advance, so I don't know if this code makes sense anymore.

fluffle: There's an implicit behavioural contract inherent in the interface, so it's not the job of this code to _"know the total number of hits in advance"_ any more. People providing their own implementations of Rater are expected to return a value from Hits() that is meaningful for their own requirements in their own context.

That could be as simple as always returning 0 because they never expect the attack to stop, or have other side-channels for stopping it. Many mathematical functions suitable for varying the attack rate can also be integrated over time to determine the total number of hits. For example, the implementation I have uses a sine wave to create a hit rate that varies between a minimum and a maximum over time.

tsenart:
I'm sorry I haven't had the time to sit down and review this properly. I'll make time for it this next weekend.

tsenart:
@fluffle: Thank you so much for this work! It's been asked over the years and I never managed to put enough time to make it happen.

When reviewing this PR, I thought that the `Rater` interface could be reduced to make it easier to implement but wasn't quite sure how. I checked out your branch locally and started playing around with different ideas. What came out is here for you to review: https://github.com/tsenart/vegeta/pull/388

Let me know what you think! Also, if you have other Rater implementations you'd like to contribute, that'd be great to validate the interface design.

tsenart:
Also @LasTshaMAN, I really appreciate the reviews you made. Thank you for helping!

fluffle:
I'll pull my sine rater into this PR as an alternative. I could probably hack together a few other simple ones too if it'd help.

fluffle:
Done. I'll port it to #388 now :-)

lib/attack.go
fluffle: This turned out to be my bad, my goimports binary was _ancient_. I only realised when it choked on the type alias :-(

fluffle:
Closing this as it's superseded by #400.

Issue #383, opened 2019-03-27
Proposal: Use content-length header instead of body for bytes in


Teddy-Schmitz:
#### Proposal

Use `response.ContentLength` in determining the number bytes in of a response rather then `len(response.Body)`

#### Background

I noticed that my bytes in was zero if I set the body limit to 0.  I don't want to read in the body but I do want to know the size of it.
To preserve previous behaviour we could do this only if `MaxBody` is set.

#### Workarounds
Don't use the `MaxBody` option

tsenart:
This would not work if the server sends data in chunked encoding, but in that case, it should be set to `-1` as per documentation: 

```
 // ContentLength records the length of the associated content. The
        // value -1 indicates that the length is unknown. Unless Request.Method
        // is "HEAD", values >= 0 indicate that the given number of bytes may
        // be read from Body.
        ContentLength int64
```

PR welcome.

Teddy-Schmitz:
Great thanks, PR #384 ready for review

PR #384, opened 2019-03-27
Use response content length to calculate bytes in

Teddy-Schmitz:
#### Background
Closes issue #383
If available use content length as the number of bytes otherwise fallback to length of body.

### DIFF
--- a/lib/attack.go
+++ b/lib/attack.go
@@ -328,6 +328,10 @@ func (a *Attacker) hit(tr Targeter, name string) *Result {
 	res.Latency = time.Since(res.Timestamp)
 	res.BytesIn = uint64(len(res.Body))
 
+	if r.ContentLength != -1 {
+		res.BytesIn = uint64(r.ContentLength)
+	}
+
 	if req.ContentLength != -1 {
 		res.BytesOut = uint64(req.ContentLength)
 	}


lib/attack.go
```diff
@@ -328,6 +328,10 @@ func (a *Attacker) hit(tr Targeter, name string) *Result {
 	res.Latency = time.Since(res.Timestamp)
 	res.BytesIn = uint64(len(res.Body))
 
+	if r.ContentLength != -1 {
+		res.BytesIn = uint64(r.ContentLength)
```
tsenart: Hmm, this doesn't look right. `BytesIn` is supposed to say how many actual bytes were read from the response, not how many bytes could be read.

Teddy-Schmitz: To me bytes in should be the size of the body regardless of whether you read it all in or not.  I would like to get the true size of the body even if I limit it with `MaxBody`.

Another option would be to use content length only if `MaxBody` is set.  An additional option would be to add a new field to represent this number and leave bytes in as is.

Let me know what your thoughts are!

tsenart: > Another option would be to use content length only if MaxBody is set. An additional option would be to add a new field to represent this number and leave bytes in as is.

This sounds like the correct option to me.

PR #385, opened 2019-03-28
Add response content length to result

Teddy-Schmitz:
#### Background

Closes issue #383

Something I wasn't sure about was what to do if Content-Length is -1 when computing metrics.  What I did for now was if it wasn't -1 add it to the total otherwise ignore it and have it be a `ByteMetric` the same as `BytesIn` and `BytesOut`  but I imagine this could give confusing results if someone is targetting multiple URLs and one returns the content length while the other doesn't.  I open to suggestions on what would be a better way to handle this.


lib/results.go
```diff
@@ -200,20 +203,24 @@ func NewCSVDecoder(rd io.Reader) Decoder {
...
 		}
 
-		r.Error = rec[5]
-		r.Body, err = base64.StdEncoding.DecodeString(rec[6])
+		if r.ContentLength, err = strconv.ParseInt(rec[5], 10, 64); err != nil {
+			return err
+		}
+
+		r.Error = rec[6]
+		r.Body, err = base64.StdEncoding.DecodeString(rec[7])
```
golangcibot: ineffectual assignment to `err` (from `ineffassign`)

Teddy-Schmitz: not sure why this came up as this is the same as previous code, but i added an error check here

lib/reporters.go
```diff
@@ -59,6 +59,7 @@ func NewTextReporter(m *Metrics) Reporter {
 		"Latencies\t[mean, 50, 95, 99, max]\t%s, %s, %s, %s, %s\n" +
 		"Bytes In\t[total, mean]\t%d, %.2f\n" +
 		"Bytes Out\t[total, mean]\t%d, %.2f\n" +
+		"Content-Length\t[total, mean]\t%d, %.2f\n" +
```
tsenart: It may be confusing to users to read this in this report together with the other byte metrics. What do you think about calling this `BodySize` instead? Do you think it'd make it more self-documenting.

lib/results.go
```diff
@@ -200,20 +203,26 @@ func NewCSVDecoder(rd io.Reader) Decoder {
...
 			return err
 		}
 
 		return err
 	}
 }
 
-// NewJSONEncoder returns an Encoder that dumps the given *Results as a JSON
-// object.
+//NewJSONEncoder returns an Encoder that dumps the given *Results as a JSON
```
tsenart: This comment is mis-formatted.

lib/reporters.go
Teddy-Schmitz: I agree I think `BodySize` would make it more clear and self-documenting.  If someone was curious how this number was found they can look at the code to see it came from content-length and body size is more descriptive.

tsenart: I just realized that BodySize or ContentLength could refer to outgoing body or incoming.

So we should disambiguate that name to something like `ResponseBodySize` (or if you think of a better name, please suggest).

tsenart:
This change also needs the README to be updated accordingly.

lib/reporters.go
Teddy-Schmitz: does `ResponseBodyLength` gives more of a clue that it comes from `content-length` ?

tsenart: Works for me.

Teddy-Schmitz: this is done

Teddy-Schmitz:
readme updated

README.md
```diff
@@ -378,13 +378,14 @@ Examples:
...
-Bytes In      [total, mean]             3714690, 3095.57
-Bytes Out     [total, mean]             0, 0.00
-Success       [ratio]                   55.42%
-Status Codes  [code:count]              0:535  200:665
+Requests                  [total, rate]             1200, 120.00
+Duration                  [total, attack, wait]     10.094965987s, 9.949883921s, 145.082066ms
+Latencies                 [mean, 50, 95, 99, max]   113.172398ms, 108.272568ms, 140.18235ms, 247.771566ms, 264.815246ms
+Bytes In                  [total, mean]             3714690, 3095.57
+Bytes Out                 [total, mean]             0, 0.00
+Response Body Length      [total, mean]             3714690, 3095.57
```
tsenart: I think the display of this information can be improved. This label is now far wider than all the others. What do you think of this alternative?

```
Requests      [total, rate]             240, 50.21
Duration      [total, attack, wait]     4.780706265s, 4.780108709s, 597.556µs
Latencies     [mean, 50, 95, 99, max]   488.965µs, 503.043µs, 571.16µs, 597.908µs, 606.328µs
Bytes In      [total, avg, rx, rx:avg]  22800, 95.00, 22800, 95.00
Bytes Out     [total, avg, tx, tx:avg]  22800, 95.00, 0, 0.00
Success       [ratio]                   100.00%
Status Codes  [code:count]              200:240
Error Set:
```

This would require documenting the semantics of all fields in `ByteMetrics` and putting the new fields there too.

tsenart:
Bump.

Teddy-Schmitz:
That seems fine but I think its going beyond the scope of the initial PR to just add the content length.  Maybe redoing the text output should be done in a separate PR?

tsenart:
I wouldn’t want users to install vegeta while the text output isn’t well done. We should do it atomically, in one PR.

README.md
Teddy-Schmitz: We dont measure the body length for bytes out so I dont think there is a reason to add `tx and tx:avg` for that it should always match the total and mean already.

As for bytes in I can change it to have the additional fields, but you also want `mean` changed to `avg`?  Then it wouldnt match the JSON reporter and the latency field.

tsenart: > We dont measure the body length for bytes out so I dont think there is a reason to add tx and tx:avg for that it should always match the total and mean already.

Good point. We should leave them out then even if we loose the consistency.

> As for bytes in I can change it to have the additional fields, but you also want mean changed to avg? Then it wouldnt match the JSON reporter and the latency field.

I think that's OK. Mean and average have the same meaning in this context. It's a trade-off to optimise the information density and visual alignment in the text report while maintaining backwards compatibility in the JSON report.

Teddy-Schmitz: ok I modified the text report and readme on this change.  I'm not sure if I fully understood what you wanted changed so let me know if you wanted it done in another way

PR #388, opened 2019-04-06
Introduce Rater interface and FixedRater

tsenart:
This commit introduces a Rater interface which allows library users to
implement alternative attack paces besides the original fixed rate.

Based on #380

### DIFF
--- a/attack.go
+++ b/attack.go
@@ -173,7 +173,8 @@ func attack(opts *attackOpts) (err error) {
 		vegeta.MaxBody(opts.maxBody),
 	)
 
-	res := atk.Attack(tr, opts.rate, opts.duration, opts.name)
+	rater := vegeta.NewFixedRater(opts.rate, opts.duration)
+	res := atk.Attack(tr, rater, opts.name)
 	enc := vegeta.NewEncoder(out)
 	sig := make(chan os.Signal, 1)
 	signal.Notify(sig, os.Interrupt)
--- a/lib/attack.go
+++ b/lib/attack.go
@@ -71,8 +71,8 @@ func NewAttacker(opts ...func(*Attacker)) *Attacker {
 
 	a.client = http.Client{
 		Transport: &http.Transport{
-			Proxy:                 http.ProxyFromEnvironment,
-			Dial:                  a.dialer.Dial,
+			Proxy: http.ProxyFromEnvironment,
+			Dial:  a.dialer.Dial,
 			ResponseHeaderTimeout: DefaultTimeout,
 			TLSClientConfig:       DefaultTLSConfig,
 			TLSHandshakeTimeout:   10 * time.Second,
@@ -211,7 +211,19 @@ func Client(c *http.Client) func(*Attacker) {
 	return func(a *Attacker) { a.client = *c }
 }
 
-// A Rate of hits during an Attack.
+// Rater defines the rate of hits during an Attack.
+type Rater interface {
+	// Interval returns the time the attacker needs to sleep
+	// before the next hit is sent to the target.
+	Interval(began time.Time, count uint64) time.Duration
+	// Hits takes the desired attack duration and returns the
+	// number of hits sent in that duration while attacking.
+	Hits(attackDuration time.Duration) uint64
+	// IsZero returns true if the rater is zero-valued.
+	IsZero() bool
+}
+
+// Rate sends a constant rate of hits to the target.
 type Rate struct {
 	Freq int           // Frequency (number of occurrences) per ...
 	Per  time.Duration // Time unit, usually 1s
@@ -222,11 +234,40 @@ func (r Rate) IsZero() bool {
 	return r.Freq == 0 || r.Per == 0
 }
 
+// Interval implements part of the Rater interface. It calculates the time
+// between each hit from r.Per / r.Freq and multiplies this by the count of
+// hits already elapsed to determine the absolute time when the next hit should
+// occur. It returns the Duration until that time.
+func (r Rate) Interval(began time.Time, count uint64) time.Duration {
+	return r.interval(began, time.Now(), count)
+}
+
+func (r Rate) interval(began, now time.Time, count uint64) time.Duration {
+	delta := time.Duration(count * uint64(r.Per.Nanoseconds()/int64(r.Freq)))
+	return began.Add(delta).Sub(now)
+}
+
+// Hits implements part of the Rater interface. It returns the number of hits
+// the attacker is expected to send when applying this Rate over the provided
+// Duration.
+func (r Rate) Hits(du time.Duration) uint64 {
+	if du == 0 || r.IsZero() {
+		return 0
+	}
+	return uint64(du) / (uint64(r.Per.Nanoseconds() / int64(r.Freq)))
+}
+
+// String returns a pretty-printed description of the rate, e.g.:
+//   Rate{1 hits/1s} for Rate{Freq:1, Per: time.Second}
+func (r Rate) String() string {
+	return fmt.Sprintf("Rate{%d hits/%s}", r.Freq, r.Per)
+}
+
 // Attack reads its Targets from the passed Targeter and attacks them at
 // the rate specified for the given duration. When the duration is zero the attack
 // runs until Stop is called. Results are sent to the returned channel as soon
 // as they arrive and will have their Attack field set to the given name.
-func (a *Attacker) Attack(tr Targeter, r Rate, du time.Duration, name string) <-chan *Result {
+func (a *Attacker) Attack(tr Targeter, r Rater, du time.Duration, name string) <-chan *Result {
 	var workers sync.WaitGroup
 	results := make(chan *Result)
 	ticks := make(chan uint64)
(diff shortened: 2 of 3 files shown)


lib/attack.go
```diff
@@ -211,7 +211,62 @@ func Client(c *http.Client) func(*Attacker) {
...
+// Wait implements the Rater interface.
+func (f RaterFunc) Wait(began, now time.Time) time.Duration {
+	return f(began, now)
+}
+
+// A FixedRater is a Rater that produces hits at a fixed rate
+// for a given duration. It is not safe for concurrent use.
+type FixedRater struct {
+	rate  Rate
+	du    time.Duration
```
golangcibot: `du` is unused (from `structcheck`)

tsenart:
Please take a look @fluffle and @LasTshaMAN.

lib/attack_test.go
```diff
@@ -356,3 +356,45 @@ func TestClient(t *testing.T) {
...
+		{1, time.Second, 10, 0, time.Second, 9 * time.Second},
+		// 1 hit/sec, 10 hits sent, 10s elapsed => 0s until next hit
+		{1, time.Second, 10, 0, 10 * time.Second, 0},
+		// 2 hit/sec, 9 hits sent, 4.4s elapsed => 100ms until next hit
+		{2, time.Second, 9, 0, (44 * time.Second) / 10, 100 * time.Millisecond},
+	} {
+		rt := Rate{Freq: tt.freq, Per: tt.per}
+		r := NewFixedRater(rt, tt.du)
+
+		for i := 0; i < tt.hits-1; i++ {
```
LasTshaMAN: I don't quite understand why you need this loop before doing the actual call (and check) on `r.Wait(began, began.Add(tt.elapsed))` ... looks like it has something to do with `rater.hits`, but as I noted above, my understanding is -  you don't need it at all ...

lib/attack.go
```diff
@@ -223,7 +223,61 @@ func Client(c *http.Client) func(*Attacker) {
 	return func(a *Attacker) { a.client = *c }
 }
 
-// A Rate of hits during an Attack.
+// A Rater defines the rate of hits during an Attack by
+// returning the duration an Attacker should wait until hiting
+// the next Target. When a negative value time is returned,
```
LasTshaMAN: (opinionated) `When a negative value time is returned` - I think Golang has multiple number of return values to adress cases like this, we can have something like `Wait(began, now time.Time) (value time.Duration, hasNext bool)`

lib/attack.go
```diff
@@ -223,7 +223,61 @@ func Client(c *http.Client) func(*Attacker) {
...
+// an Attacker will stop.
+type Rater interface {
+	Wait(began, now time.Time) time.Duration
```
LasTshaMAN: Function signature looks weird and counter-intuitive to me ... why do we need `began` and `now` parameters in order to be able to return `time.Duration` on each consequtive call ?

I think we leak implementation details here ...
What `Rater` basically is - an integer (time.Duration if we want to be more specific) sequence generator, independent of the concept of time.

lib/attack.go
```diff
@@ -223,7 +223,61 @@ func Client(c *http.Client) func(*Attacker) {
...
+	rate  Rate
+	hits  uint64
+	count uint64
+}
+
+// NewFixedRater returns a new FixedRater with the given Rate and duration.
+// When duration is zero, the rater will never stop.
+func NewFixedRater(rate Rate, du time.Duration) *FixedRater {
+	r := FixedRater{rate: rate}
+	if du != 0 && !rate.IsZero() {
```
fluffle: If `rate.IsZero` returns true, this should probably return an error and a nil `*FixedRater`. Otherwise the implementation of `Wait` as it is now could panic with integer-divide-by-zero if `rate.Freq` is zero.

attack.go
```diff
@@ -176,7 +176,8 @@ func attack(opts *attackOpts) (err error) {
 		vegeta.UnixSocket(opts.unixSocket),
 	)
 
-	res := atk.Attack(tr, opts.rate, opts.duration, opts.name)
+	rater := vegeta.NewFixedRater(opts.rate, opts.duration)
+	res := atk.Attack(tr, rater, opts.name)
```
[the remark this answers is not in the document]
fluffle: Drat. Preserving backwards compatibility isn't possible when pushing the responsibility for counting hits to the `Rater`, because the Rate struct needs to have a pointer receiver for its methods. That's why it's wrapped with the `FixedRater` struct here.

lib/attack.go
LasTshaMAN: This also point out a dent in tests coverage ;)

lib/attack.go
fluffle: Having messed around with this some, I think the first alternative is actually better. The approach I've converged on is:

*   `Attack` should be responsible for terminating the attack after any non-zero duration has elapsed.
    *    This doesn't seem like the job of code that is supposed to be pacing hits.
*   `Attack` should track both the elapsed time of the attack and the number of hits sent.
    *    It should provide the `Pacer` (previously `Rater`) with both of these.
    *    This means the `Pacer` doesn't necessarily need to maintain internal state, so it's possible to preserve backwards compatibility with the current `Rate` struct.
    *    The Pacer is free to ignore these if its implementation doesn't need them.
*   `Pacer` still needs some way of stopping the attack that's not necessarily duration based.
    *   This allows for control logic to e.g. abort load tests when errors / latency are outside acceptable thresholds.

Thus, I've arrived at:

```go
type Pacer interface {
	Pace(elapsedTime time.Duration, elapsedHits uint64) (sleep time.Duration, stop bool)
}
```

lib/attack.go
LasTshaMAN: You are right, I haven't thought it through.

lib/attack.go
LasTshaMAN: Yeah, I also think that `Pacer` shouldn't be responsible for `the elapsed time of the attack and the number of hits sent`. 

Also, in my opinion the name `Pacer` reflects the intention better than `Rater`.

lib/attack.go
```diff
@@ -223,7 +223,61 @@ func Client(c *http.Client) func(*Attacker) {
...
+
+// Wait is an utility method that calls the rater function itself.
+// It's meant to aid readability.
+func (r *FixedRater) Wait(began, now time.Time) time.Duration {
+	if r.hits != 0 && r.count == r.hits {
+		return -1
+	}
+
+	r.count++
+	interval := uint64(r.rate.Per.Nanoseconds() / int64(r.rate.Freq))
```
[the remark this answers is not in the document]
tsenart: That makes sense. I'll implement it.

lib/attack.go
tsenart: Agree with everything here. This is a great example of collective design! Happy about this outcome.

```
type Pacer interface {
	Pace(elapsedTime time.Duration, elapsedHits uint64) (sleep time.Duration, stop bool)
}
```

lib/attack.go
tsenart: > For testing purposes it's simple to have a separate internal implementation that takes a now parameter, as I did in #380.

Yeah, it was for testing purposes. I don't particularly mind using an API that makes me call time.Now() before, but I understand that's not everyone's cup of tea. So I'm cool with making it more ergonomic.

lib/attack.go
```diff
@@ -223,7 +223,61 @@ func Client(c *http.Client) func(*Attacker) {
...
+// returning the duration an Attacker should wait until hiting
+// the next Target. When a negative value time is returned,
+// an Attacker will stop.
+type Rater interface {
+	Wait(began, now time.Time) time.Duration
+}
+
+// A RaterFunc is a function adapter type that implements the
+// Rater interface.
+type RaterFunc func(began, now time.Time) time.Duration
```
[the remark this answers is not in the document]
tsenart: >  Since we're expecting Rater implementors to track the hit count now, I would expect most implementations to require a struct or at least a uint64 for this anyway.

I believe with the `Pacer` design this isn't relevant anymore, but here's how you'd maintain state with a function closure.

```go
// NewLinearRater returns a Rater that increases it's wait time by
// the given duration every time it is called.
func NewLinearRater(step time.Duration) Rater {
    var count int64
    return RaterFunc(func(began, now time.Time) time.Duration {
        count++
        return began.Add(step * time.Duration(count))
    })
}
```

lib/attack.go
```diff
@@ -223,7 +223,61 @@ func Client(c *http.Client) func(*Attacker) {
...
+
+// Wait implements the Rater interface.
+func (f RaterFunc) Wait(began, now time.Time) time.Duration {
+	return f(began, now)
+}
+
+// A FixedRater is a Rater that produces hits at a fixed rate
+// for a given duration. It is not safe for concurrent use.
```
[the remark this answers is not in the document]
tsenart: It could easily be made concurrent safe, but I couldn't see use cases that'd justify that need. Do you think it'd be useful, for instance, to use the same `FixedRater` in two concurrent attacks in separate `Attackers`?

attack.go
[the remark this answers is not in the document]
tsenart: I'm on-board with trying to minimize breaking backwards compatibility, despite Go modules and vendoring making this less of a problem than it was in the past.

lib/attack.go
```diff
@@ -223,7 +223,61 @@ func Client(c *http.Client) func(*Attacker) {
...
+		r.hits = uint64(du) /
+			(uint64(rate.Per.Nanoseconds() / int64(rate.Freq)))
+	}
+	return &r
+}
+
+// Wait is an utility method that calls the rater function itself.
+// It's meant to aid readability.
+func (r *FixedRater) Wait(began, now time.Time) time.Duration {
+	if r.hits != 0 && r.count == r.hits {
```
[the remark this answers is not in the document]
tsenart: I believe @fluffle's comments and ensuing `Pacer` proposal have shed some light on this so I'm resolving this comment. Let me know if you'd like more context otherwise.

lib/attack.go
tsenart: Good catch. What do you think of using a zero valued Rate as meaning no pacing at all? Just go as fast as you can.

PR #389, opened 2019-04-15
SineRater on top of rater.

fluffle:
#### Background

#380 and #388 

Not expecting this to be merged as-is, just demonstrating #380's SineRate on top of the Rater interface from #388.

@tsenart @LasTshaMAN


tsenart:
Let's go with Pacer. Thanks for the extra effort.

PR #390, opened 2019-04-15
Introduce Pacer (née Rater) interface.

fluffle:
Again not expecting this to be merged, but this is the direction I'd take things in.


lib/pacer.go
```diff
@@ -0,0 +1,170 @@
...
+	"math"
+	"time"
+)
+
+// A Pacer defines the rate of hits during an Attack by
+// returning the duration an Attacker should wait until hitting
+// the next Target. If the second return value is true, the
+// attack will terminate.
+type Pacer interface {
+	Pace(elapsedTime time.Duration, elapsedHits uint64) (sleep time.Duration, stop bool)
```
tsenart:
```suggestion
	Pace(elapsed time.Duration, hits uint64) (wait time.Duration, stop bool)
```

lib/pacer.go
```diff
@@ -0,0 +1,170 @@
...
+	return delta - elapsedTime, false
+}
+
+// String returns a pretty-printed description of the rate, e.g.:
+//   Rate{1 hits/1s} for Rate{Freq:1, Per: time.Second}
+func (r Rate) String() string {
+	return fmt.Sprintf("Rate{%d hits/%s}", r.Freq, r.Per)
+}
+
+func (r Rate) HitsPerNs() float64 {
```
tsenart: Please add Go doc.

lib/pacer.go
```diff
@@ -0,0 +1,170 @@
...
+// SinePacer is a Pacer that describes attack request rates with the equation:
+//     R = MA sin(O+(2𝛑/P)t)
+// Where:
+//   R = Instantaneous attack rate at elapsed time t, hits per nanosecond
+//   M = Mean attack rate over period P, sp.Mean, hits per nanosecond
+//   A = Amplitude of sine wave, sp.Amp, hits per nanosecond
+//   O = Offset of sine wave, sp.StartAt, radians
+//   P = Period of sine wave, sp.Period, nanoseconds
+//   t = Elapsed time since attack, nanoseconds
+// The attack rate (sp.HitsPerNs) is described by the equation:
```
tsenart: This equation seems to be missing.

lib/pacer.go
```diff
@@ -0,0 +1,170 @@
...
+//
+// This equation is integrated with respect to time to derive the expected
+// number of hits served at time t after the attack began:
+//     H = Mt - (AP/2𝛑)cos(O+(2𝛑/P)t) + (AP/2𝛑)cos(O)
+// Where:
+//   H = Total number of hits triggered during t
+type SinePacer struct {
```
tsenart: Amazing docs! Do you think we could include a little ASCII art sine wave describing the shape of the attack rate for the less mathematically inclined (like me)?.

lib/pacer.go
```diff
@@ -0,0 +1,170 @@
...
+	StartAt float64
+}
+
+var _ Pacer = SinePacer{}
+
+func (sp SinePacer) String() string {
+	return fmt.Sprintf("Sine{%s ± %s / %s, offset %g𝛑}", sp.Mean, sp.Amp, sp.Period, sp.StartAt/math.Pi)
+}
+
+// Pace returns the Duration until the next hit should be sent,
```
tsenart:
```suggestion
// Pace returns the Duration to wait until the next hit is sent,
```

lib/pacer.go
```diff
@@ -0,0 +1,170 @@
...
+	// The amplitude of the sine wave in freq-per-Duration,
+	// MUST NOT BE EQUAL TO OR LARGER THAN MEAN
+	Amp Rate
+	// The offset, in radians, for the sine wave at t=0.
+	StartAt float64
+}
+
+var _ Pacer = SinePacer{}
+
+func (sp SinePacer) String() string {
```
tsenart: Go doc please.

lib/pacer.go
```diff
@@ -0,0 +1,170 @@
...
+	return fmt.Sprintf("Rate{%d hits/%s}", r.Freq, r.Per)
+}
+
+func (r Rate) HitsPerNs() float64 {
+	return float64(r.Freq) / float64(r.Per)
+}
+
+var _ Pacer = Rate{}
+
+const (
```
tsenart: What are these constants used for? It'd be good to explain in docs.

tsenart: Each one individually.

lib/pacer_test.go
```diff
@@ -0,0 +1,148 @@
...
+
+var quarterPeriods = map[string]float64{
+	"MeanUp":   MeanUp,
+	"Peak":     Peak,
+	"MeanDown": MeanDown,
+	"Trough":   Trough,
+}
+
+type sineTest struct {
+	p, m, a int
```
tsenart: Please give these more descriptive names and document how they're used. It's hard for me to understand otherwise.

lib/pacer_test.go
```diff
@@ -0,0 +1,148 @@
...
+			wait, stop := p.Pace(test.et, test.c)
+			if wait != test.wait || stop != test.stop {
+				t.Errorf("%d(%s): wait(%v) = (%v, %v), want (%v, %v)",
+					i, name, test.et, wait, stop, test.wait, test.stop)
+			}
+		}
+	}
+}
+
+// No idea how to test interval without hard-coding a bunch of numbers.
```
tsenart: Neither would I :D

lib/pacer.go
```diff
@@ -0,0 +1,170 @@
...
+// number of hits served at time t after the attack began:
+//     H = Mt - (AP/2𝛑)cos(O+(2𝛑/P)t) + (AP/2𝛑)cos(O)
+// Where:
+//   H = Total number of hits triggered during t
+type SinePacer struct {
+	// The period of the sine wave, e.g. 20*time.Minute
+	// MUST BE > 0
+	Period time.Duration
+	// The mid-point of the sine wave in freq-per-Duration,
+	// MUST BE > 0
```
tsenart: Visual cues of how each of these parameters affect the sine wave would be great!

tsenart:
Magic! I got the Mean and Amp parameters the other way around initially, so perhaps we can improve the docs. Otherwise, how cool is this? :dancer: 
Peek 2019-04-21 15-17

lib/attack.go
```diff
@@ -313,14 +243,16 @@ func (a *Attacker) Attack(tr Targeter, r Rater, name string) <-chan *Result {
 		defer close(ticks)
 		began := time.Now()
 		for {
-			wait := r.Wait(began, time.Now())
-			if wait < 0 {
+			elapsed := time.Since(began)
+			wait, stop := p.Pace(elapsed, count)
+			if stop || du > 0 && elapsed > du {
```
LasTshaMAN: I think it would be clearer if we check `du > 0` just once - the very first thing we do in `Attack` function - and return if this condition is not satisfied.

lib/pacer.go
```diff
@@ -0,0 +1,170 @@
...
+		}
+		nextHitIn = time.Duration(float64(nextHitIn) / (hitsAtGuess - float64(elapsedHits)))
+	}
+	return nextHitIn, false
+}
+
+// AmpHits returns AP/2𝛑, which is the number of hits added or subtracted
+// from the Mean due to the Amplitude over a quarter of the Period,
+// i.e. from 0 → 𝛑/2 radians
+func (sp SinePacer) AmpHits() float64 {
```
LasTshaMAN: Do helper-methods like these need to be exported (starting with capital letter) ? If we need them for implementation only, then it would be clearer for the end user if we hide them from him.

lib/attack.go
```diff
@@ -223,85 +223,15 @@ func Client(c *http.Client) func(*Attacker) {
...
-}
-
-// String returns a pretty-printed description of the rate, e.g.:
-//   Rate{1 hits/1s} for Rate{Freq:1, Per: time.Second}
-func (r Rate) String() string {
-	return fmt.Sprintf("Rate{%d hits/%s}", r.Freq, r.Per)
-}
-
 // Attack reads its Targets from the passed Targeter and attacks them at
 // the rate produced by the given Rater. Results are sent to the returned
```
LasTshaMAN:
```suggestion
 // the rate produced by the given Pacer. Results are sent to the returned
```

lib/pacer.go
```diff
@@ -0,0 +1,170 @@
...
+// A PacerFunc is a function adapter type that implements the
+// Pacer interface.
+type PacerFunc func(time.Duration, uint64) (time.Duration, bool)
+
+// Pace implements the Pacer interface.
+func (f PacerFunc) Pace(elapsedTime time.Duration, elapsedHits uint64) (time.Duration, bool) {
+	return f(elapsedTime, elapsedHits)
+}
+
+// Rate sends a constant rate of hits to the target.
```
LasTshaMAN:
```suggestion
 // Rate defines a constant rate of hits for the target.
```

lib/pacer.go
```diff
@@ -0,0 +1,170 @@
...
+type Rate struct {
```
LasTshaMAN: How about naming it `FixedRate` or `FixedRatePacer` in light of the latest changes ?

Also, consider renaming `SinePacer` so that both implementations look similar (that is either use `Pacer` suffix for all of them or none of them)

lib/pacer_test.go
```diff
@@ -0,0 +1,148 @@
...
+
+	for i, tt := range []struct {
+		freq    int
+		per     time.Duration
+		elapsed time.Duration
+		count   uint64
+		wait    time.Duration
+		stop    bool
+	}{
+		// NOTE: Rate sleeps before sending the first hit,
```
LasTshaMAN: First off, Rate doesn't actually sleep, it just returns some `time.Duration` for the caller to do with it as he wishes. Thus this might be misleading for the reader.

Second, this is an implementation detail and is easily deduced from some of the test-cases below, so, I would say this comment is not necessary here at all.

lib/pacer_test.go
```diff
@@ -0,0 +1,148 @@
+package vegeta
+
+import (
+	"math"
+	"testing"
+	"time"
+)
+
+func TestRatePacing(t *testing.T) {
```
LasTshaMAN: What about all the edge cases ?

- zero hits per second
- 1 hit per zero seconds
- zero hits per zero seconds
- other cases when the second argument turns out to become `true`

lib/pacer_test.go
```diff
@@ -0,0 +1,148 @@
...
+	return time.Duration(1 / (float64(st.m) + float64(st.a)*math.Sin(startAt+frac*2*math.Pi)))
+}
+
+func TestSinePacerHits(t *testing.T) {
+	tests := []sineTest{
+		{20 * 60, 100, 90},
+		{60, 1000, 10},
+		{1, 10, 7},
+		{1, 1, 0},
+		// These test cases failed with off-by-one errors before applying
```
LasTshaMAN: I think this comment belongs to the place where you apply this rounding (to clarify why it is done so) rather than here.

lib/pacer.go
LasTshaMAN:
```suggestion
	Pace(elapsed time.Duration, hitsDone uint64) (wait time.Duration, stop bool)
```
would also be a good alternative

lib/pacer_test.go
```diff
@@ -0,0 +1,148 @@
...
+				t.Errorf("%d(%s): hits after 3/4 period = %d, want %d", i, name, got, want)
+			}
+			if got, want := sr.Hits(sr.Period), test.Hits(1, sa); got != want {
+				t.Errorf("%d(%s): hits after full period = %d, want %d", i, name, got, want)
+			}
+		}
+	}
+}
+
+func TestSinePacerFlat(t *testing.T) {
```
LasTshaMAN: Don't forget edge cases also

lib/pacer_test.go
```diff
@@ -0,0 +1,148 @@
...
+		// These test cases failed with off-by-one errors before applying
+		// math.Round in Hits, due to floating-point maths differences.
+		{1e6, 10, 7},
+		{60, 1000, 999},
+	}
+
+	for i, test := range tests {
+		for name, sa := range quarterPeriods {
+			sr := test.Pacer(sa)
+			if got, want := sr.Hits(sr.Period/4), test.Hits(0.25, sa); got != want {
```
LasTshaMAN: What about edge case `sr.Hits(0)` ?

lib/pacer_test.go
```diff
@@ -0,0 +1,148 @@
...
+		stop bool
+	}{
+		{0, 0, time.Second, false},
+		{0, 1, 2 * time.Second, false},
+		{time.Second / 100, 0, 99 * time.Second / 100, false},
+		{time.Second / 2, 0, time.Second / 2, false},
+		{64 * time.Second / 100, 0, 36 * time.Second / 100, false},
+		// Has an off-by-one I can't round away nicely because it's
+		// due to expectedHits being 0.9900000000000001. Ugh floats.
+		// {99 * time.Second / 100, 0, time.Second / 100},
```
LasTshaMAN: What about comparing floats "to a certain degree" (say, using https://github.com/google/go-cmp/tree/master/cmp) ?

I think, it is the only reasonable approach ...

lib/pacer_test.go
LasTshaMAN: also, consider using not-exact comparison for floats

lib/pacer_test.go
LasTshaMAN: Property-based testing could help us out here (look up some presentations on youtube).

The basic idea is:
- instead of using concrete examples to test our code we figure out some properties for our code that absolutely must hold, and we describe these properties in tests.

Here are some examples ... suppose I have time interval of a certain predefined `length`, I also have fixed rate pacer (with the rate, say, `freq`/`per`):
- then I know that for any `freq` and `per` (we can generate them randomly) and any `lenght` if I call `Pace(length, x)` where `x = any from 0 to (freq * (length / per) - 1)` I should get `(0, false)` back, meaning pacer has fallen behind and must return no "wait time" at all;
- then I know that for any `freq` and `per` (we can generate them randomly) and any `lenght` if I call `Pace(length, x)` where `x = any from (freq * (length / per) + 1) to MAX_INT` I should get `(not zero, false)` back, meaning pacer is ahead and has no right to return 0 for "wait time"
- then I know that while I'm pacing some interval I cannot accumulate "wait time" greater than the length for that interval
- etc.

lib/pacer.go
```diff
@@ -0,0 +1,170 @@
...
+
+// Pace determines the length of time to sleep until the next hit is sent.
+func (r Rate) Pace(elapsedTime time.Duration, elapsedHits uint64) (time.Duration, bool) {
+	if r.Per <= 0 || r.Freq <= 0 {
+		// If Rate configuration is invalid, stop the attack.
+		return 0, true
+	}
+	interval := uint64(r.Per.Nanoseconds() / int64(r.Freq))
+	delta := time.Duration((elapsedHits + 1) * interval)
+	// Zero or negative durations cause time.Sleep to return immediately.
```
LasTshaMAN: (opinionated) I would simplify this to zero, avoiding negative cases as they don't make much sense for the caller.

lib/pacer.go
```diff
@@ -0,0 +1,170 @@
...
+}
+
+// Pace determines the length of time to sleep until the next hit is sent.
+func (r Rate) Pace(elapsedTime time.Duration, elapsedHits uint64) (time.Duration, bool) {
+	if r.Per <= 0 || r.Freq <= 0 {
+		// If Rate configuration is invalid, stop the attack.
+		return 0, true
+	}
+	interval := uint64(r.Per.Nanoseconds() / int64(r.Freq))
+	delta := time.Duration((elapsedHits + 1) * interval)
```
LasTshaMAN: watch out for `uint64` overflow - stumbled on it while testing

lib/pacer_test.go
LasTshaMAN: There probably exist some testing frameworks for property-based testing, but here is fast&dirty way to test the properties I described above:

```go
func TestFixedRatePacerPropertyBased(t *testing.T) {
	rand.Seed(time.Now().UnixNano())

	t.Run("when pacer has fallen behind it must return zero wait time", func(t *testing.T) {
		for i := 0; i < 100; i++ {
			freq := 1 + rand.Intn(math.MaxInt32-1)
			per := time.Duration(1 + rand.Int63n(math.MaxInt64-1))
			// time interval length
			length := time.Duration((int64(per)/int64(freq) + 1) + rand.Int63n(10000))

			expectedHitsForInterval := float64(freq) * (float64(length) / float64(per))
			hitsDone := uint64(rand.Int63n(int64(math.Floor(expectedHitsForInterval))))

			r := &Rate{Freq: freq, Per: per}
			wait, stop := r.Pace(length, hitsDone)
			if stop == true {
				t.Errorf("freq=%v per=%v length=%v, stop: exp = %v, act = %v", freq, per, length, false, stop)
			}
			if wait > 0 {
				t.Errorf("freq=%v per=%v length=%v, wait: exp = zero or negative number, act = %v", freq, per, length, wait)
			}
		}
	})
	t.Run("when pacer is ahead it must return positive wait time", func(t *testing.T) {
		for i := 0; i < 100; i++ {
			freq := 1 + rand.Intn(math.MaxInt32-1)
			per := time.Duration(1 + rand.Int63n(math.MaxInt64-1))
			// time interval length
			length := time.Duration((int64(per)/int64(freq) + 1) + rand.Int63n(10000))

			expectedHitsForInterval := float64(freq) * (float64(length) / float64(per))
			boundary := int64(math.Ceil(expectedHitsForInterval))
			hitsDone := uint64(boundary + rand.Int63n(math.MaxInt64-boundary))
			//hitsDone := uint64(boundary + 22222222)

			r := &Rate{Freq: freq, Per: per}
			wait, stop := r.Pace(length, hitsDone)
			if stop == true {
				t.Errorf("freq=%v per=%v length=%v, stop: exp = %v, act = %v", freq, per, length, false, stop)
			}
			if wait <= 0 {
				t.Errorf("freq=%v per=%v length=%v, wait: exp = positive number, act = %v", freq, per, length, wait)
			}
		}
	})
	t.Run("pacing some interval I cannot accumulate wait time greater than the length for that interval", func(t *testing.T) {
		for i := 0; i < 100; i++ {
			freq := 1 + rand.Intn(math.MaxInt32-1)
			per := time.Duration(1 + rand.Int63n(math.MaxInt64-1))
			// time interval length
			length := time.Duration((int64(per)/int64(freq) + 1) + rand.Int63n(10000))

			elapsed := time.Duration(0)
			hits := uint64(0)

			maxHitsExp := uint64(math.Ceil((float64(length) / float64(per)) * float64(freq)))

			r := &Rate{Freq: freq, Per: per}
			for hits < maxHitsExp {
				wait, stop := r.Pace(length, hits)
				if stop == true {
					t.Errorf("freq=%v per=%v length=%v, stop: exp = %v, act = %v", freq, per, length, false, stop)
				}
				elapsed += wait
				hits++
			}

			if elapsed > length {
				t.Errorf("freq=%v per=%v length=%v, wait: elapsed = %v", freq, per, length, elapsed)
			}
		}
	})
}
```

Some of them fails probably because of overflows (haven't had time to look). Also I'd probably reduce some code duplication in these tests if they get through to the final PR.

LasTshaMAN: If you like this approach, I coult tidy things up and write some more property-based tests.

lib/pacer.go
fluffle: Rate needed to remain named Rate for backwards compatibility. We could use a [type alias](https://go.googlesource.com/proposal/+/master/design/18130-type-alias.md), I suppose, if we're willing to restrict users of the library to go1.9+. WDYT?

lib/pacer.go
fluffle: This comment is not here for the caller, it is here for the maintainer, to explain why I'm not bothering to check whether `delta - elapsedTime` is negative and clamp to zero :-)

lib/pacer.go
fluffle: :O

The docs say time.Duration is good for durations of up to ~290 years. Practically speaking, if you are anywhere near the overflow point, Pace is not going to be doing anything useful. But I can add a check for e.g. `INT_MAX / (elapsedHits + 1) > interval` to detect overflows and return `0, true` in this case if you think it's necessary.

lib/pacer.go
fluffle: Will do. Sorry, I just copied over the internal version and lightly edited it, didn't really do the proper due diligence.

lib/pacer.go
fluffle: Will do. They're useful constant values for the `StartAt` `SinePacer` struct field.

lib/pacer.go
fluffle: Ooops, overzealous deletion. Good catch, thanks :-)

lib/pacer.go
fluffle: Eeeerrh. That is an unexpected request ;-)

I will see what I can come up with, my ascii art skillz are pretty minimal...

lib/pacer.go
fluffle: Will do.

lib/pacer.go
fluffle: Will do.

lib/pacer.go
fluffle: Good question.

Arguably not, I suppose. I exported them just in case they might be useful but I've not yet found them to be. I'll make them private.

lib/pacer_test.go
fluffle: I just copied these tests over from the other file, I agree they need some additions. I'm happy to augment them with more tests.

lib/pacer_test.go
fluffle: OK, fair point, I'll remove this.

lib/pacer_test.go
fluffle: Sorry. I wrote these tests quickly, and I'm very lazy when it comes to testing :-(

lib/pacer_test.go
fluffle: There's a comment there too (L113 of pacer.go). Note that these aren't floats, the problem was that the integer truncation in pacer.go occasionally truncated to unexpected values.

lib/pacer_test.go
fluffle: quarterPeriods contains MeanUp which is 0.

lib/pacer_test.go
fluffle: These tests are pretty hacky, I wrote them quickly but relied on "manual integration testing" (i.e. push it and see if it does the right thing) to validate things mostly. I am a terrible person ;-)

I agree that for the final PR they need a lot of work.

lib/pacer_test.go
fluffle: Thank you, that's absolutely what I should have been doing :-)

lib/pacer_test.go
fluffle: Interesting! Does it extend nicely to variable rates? The first two look like they should as long as you can determine expectedHits from elapsedTime for your Pacer, but I can't tell as easily for the third one.

I'd definitely be grateful for a cleaned-up test suite that you could just feed pacer implementations to.

lib/pacer.go
LasTshaMAN: Oops! Forgot about backward-compatibility. 

Using type alias looks reasonable to me. That's for @tsenart to decide :)

lib/pacer.go
LasTshaMAN: What I mean was - I think it would be better to never return negative number from this function. Although `time.Duration` allows for this outcome, I think logically it doesn't make any sense and requires that caller be wary of it ...

If we always return `0` instead of negatives we might help the caller to avoid doing "something stupid". 

Testing it also will be easier (tests are what actually forced me to think about this in the first place, and most of the times testing-first leads to better design ...)

Just trying to simplify things wherever possible :)

lib/pacer.go
LasTshaMAN: > Practically speaking, if you are anywhere near the overflow point, Pace is not going to be doing anything useful

You are right, I was testing the full range of input arguments, and it is more a thing of a taste ...

I think tests should drive such decisions (whether to add check or not), as it would be simpler to write them (provided that you wrote good quality tests :) ).

lib/pacer_test.go
LasTshaMAN: > but I can't tell as easily for the third one

I think it does ... we can run it several times and be pretty sure whether or not :)

> Does it extend nicely to variable rates?

Some properties will be true for all `Pacer` implementations, others may be specific to only few, it depends on what properties we will come up with to validate our code :)

The idea of creating properties that every `Pacer` must satisfy feels good to me as we will be able to test new `Pacer` implementation just by inserting them in our test-suite.

If you any other properties (other that 3 listed above) come to your mind - let me know :)

LasTshaMAN:
> And if I wanted to take up @LasTshaMAN on their offer of writing tests for me (yay!), what's the best way of doing that? Should I ask them for a PR to my branch, and then merge that in? Should they send a separate PR with tests once mine has been merged?

I guess any approach will be fine by me. 

I also have a question for @tsenart :) 

Can I use 3rd party libraries for testing (e.g. https://github.com/stretchr/testify) to ease some pain of writing tests ? Go standard testing library is not bad, but, oh dear God, no one should write assert statements through `if`s ... May be there are some nice property-based testing libraries that may be used for writing such tests in cleaner manner (than what I wrote above) ...

lib/pacer.go
tsenart: Thanks. Don't sweat over it though, it's not a blocker. You may find this helpful: http://ascii.co.uk/art/sine

lib/pacer_test.go
tsenart: No problem, I understand! It's important to have the tests as maintainable as possible though.

lib/pacer_test.go
tsenart: > If you like this approach, I could tidy things up and write some more property-based tests.

> The idea of creating properties that every Pacer must satisfy feels good to me as we will be able to test new Pacer implementation just by inserting them in our test-suite.

Yes and YES! Property based testing is awesome. I use it a lot in another OSS project I maintain. Example: https://github.com/oklog/ulid/blob/master/ulid_test.go#L116

This level of contribution goes beyond my wildest dreams. Thank you both!

lib/pacer.go
tsenart: I never used type aliases before. At first glance, it seems fine. I don't mind restricting this library to go1.9+. Are there any other caveats?

As for the name, what do you think of `ConstantPacer` or `SteadyPacer`?

lib/pacer_test.go
tsenart: > I'm happy to augment them with more tests.

Covering the edge cases seems important.

lib/pacer_test.go
tsenart: @LasTshaMAN: It seems that you are quite inspired on the testing front. Would you be up to contribute this part of the work?

tsenart:
> What's the best way of polishing this up for a merge? Do you want to preserve the history of #380, #388 and this? Would you prefer me to create a new PR branched from master with a (new) linear history of logical commits? I'm happy to do either.

If you don't mind going through it, a linear history of logical commits is nice to have. But not an absolute requirement. I'm happy with a single commit too.

> And if I wanted to take up @LasTshaMAN on their offer of writing tests for me (yay!), what's the best way of doing that? Should I ask them for a PR to my branch, and then merge that in? Should they send a separate PR with tests once mine has been merged?

I think the best way would be to merge @fluffle's PR first, and then work on a follow up PR authored by @LasTshaMAN that'd implement the thorough tests and address any bugs found. This is OK, since users should be relying on semantic versions anyways, not on master.

> Can I use 3rd party libraries for testing (e.g. https://github.com/stretchr/testify) to ease some pain of writing tests ? Go standard testing library is not bad, but, oh dear God, no one should write assert statements through ifs ... Maybe there are some nice property-based testing libraries as well - that will help writing such tests in a cleaner manner (than what I wrote above) ...

Please, no 3rd party libraries unless they're reeeeeally necessary. Did you know about https://golang.org/pkg/testing/quick/? See how I use it in oklog/ulid: https://sourcegraph.com/search?q=repo:github.com/oklog/ulid+quick.Check

lib/pacer.go
LasTshaMAN: I think `SteadyPacer` might not be exact enough :) `ConstantPacer` looks great!

lib/pacer_test.go
LasTshaMAN: Yeah, I like testing things :)

Sure, I'll try my best, just let me know which form of contribution (as @fluffle mentioned below) is preferable.

LasTshaMAN: Ok, saw the comment below, a follow-up PR would be convenient for me to do. 

Let's finish up with this one and then I'll get to work on testing new `Pacer` functionality (as long as I have spare time of course, this May I'm a little busy ...).

LasTshaMAN:
> Did you know about https://golang.org/pkg/testing/quick/? See how I use it in oklog/ulid: https://sourcegraph.com/search?q=repo:github.com/oklog/ulid+quick.Check

Yeah, I saw it used a couple of times :)

> Please, no 3rd party libraries unless they're reeeeeally necessary.

Ok, it's not a deal breaker. 

Just curious, how do u define "reeeeeally necessary" ? I mean, with a nice package management system in place (`vgo` as is now, and possibly converting to `go modules`) why does it matter how many 3rd party dependencies the project has ?

I know of some problems that may result from pulling in every possible library (instead of using only core lang libraries), e.g.
- 3rd party library can disappear from the Internet (this can be solved via creating `vendor` directory to store all the dependencies locally to the target project)
- if the target project relies on some functionality broken in 3rd party library then it would be harder to fix it (as we have to update 3rd party project and pull in the fix into ours)
- ...

but I think none of these are real impediments to the development process, compared to lost advantage of reusing 3rd party code (for nice assertions, say).

fluffle:
Superseded by #400.

Issue #394, opened 2019-04-23
Proposal: Include histogram in JSON output


fxkr:
#### Proposal

Add histogram data to the `vegeta report -type json` output, e.g.:

```sh
# vegeta report -type json -hist '[0,5ms,10ms,15ms]'
  ...
  "hist": [123, 125, 40, 80],
  ...
```

I'd be happy to write the code, but I'd like to run some design decisions by you first:

* Under what key: don't care. `hist` matches CLI.

* List **(preferred)** or map. It's guaranteed ordering and no questions about the format of the key ("min", "max", "min-max") vs self-descriptiveness.

* Values in msec

* Bin config:
  * **(preferred)** Separate command line argument (`-type json -hist [0,5ms,10ms,15ms]`)
  * As a subsequent map (`-type json'{"hist":["0","5ms","10ms","15ms"]}'`)
  * As a subsequent array (`-type json'[0,5ms,10ms,15ms]'`). Similar to `hist`, but I don't like this approach because it's not extensible at all, and I don't want to have to deal with breaking changes later.

  Either way I'd default to leaving out histogram data if no config is given for it.

  Automatic logarithmic binning would be kinda neat (I love this in [bcc](https://github.com/iovisor/bcc)), but I see it as out of scope.

Do you think that would be useful?

The main concern I have is that I am not 100% sure if recording the histogram is the best idea. Having just the three percentiles that vegeta puts in the JSON right now is too coarse, but we could just output more of those instead. I wonder if you have any thoughts on recording histograms vs (many points of the) CDF?

#### Background

I want to automatically benchmark a piece of software as part of a CI/CD pipeline and record the results to track how it changes over time. I want to record more than just the percentiles because the latency distribution can be strongly multimodal.

For example, depending on the concrete test we can hit (and possibly trash) or bypass certain caches (at different layers in our software), and having the histogram lets us see that very clearly visually.

#### Workarounds

* Parse `report -type hist` human-readable output
* Use vegeta as a library

Thanks for writing vegeta!

fxkr:
I went ahead and did it the way I think it's best; it was pretty easy. Let me know if you'd prefer it to be done differently.

tsenart:
> Automatic logarithmic binning would be kinda neat (I love this in bcc), but I see it as out of scope.

I'd LOVE this, really!

PR #395, opened 2019-04-24
Include histogram in JSON output

fxkr:
#### Background

See #394

Interesting cases to test:

```shell
# Main use case
% vegeta report -type=json -hist='[0,1ms,2ms,3ms,4ms,5ms]' vegeta.bin
{"latencies": ..., "hist": [0,0,0,56,44,400], ...}

# If -hist is not given, "hist" is omitted
% vegeta report -type=json vegeta.bin

# -type=hist supports -hist too (for consistency)
% vegeta report -type=hist -hist='[0,1ms,2ms,3ms,4ms,5ms]' vegeta.bin

# Legacy -type=hist invocation continues to work unchanged
% vegeta report -type=hist'[0,1ms,2ms,3ms,4ms,5ms]' vegeta.bin

# Help should mention the new parameter
% vegeta --help
```

#### Checklist


tsenart:
Thanks for the contribution! I have some thoughts that I'd like to share and get your input on.

I've had the feeling for a long time that the histogram report is under-used and its UX and utility could be substantially improved. Your use-case hints at the same conclusion.

Similarly to how `vegeta plot` became its own sub command when it started to deviate significantly from the other "reports", I think that the introduction of this `--hist` command line argument suggests that a `vegeta hist(ogram)` sub-command is up for promotion.

My vision for this sub-command could be captured like this in its help output:

```console
Usage: vegeta histogram [options] [<file>...]

Outputs a histogram of request latencies in HTML, JSON or text format.

Arguments:
  <file>  A file with vegeta attack results encoded with one of
          the supported encodings (gob | json | csv) [default: stdin]

Options:
  --bins    How to bin request latencies. Defaults to auto
            logarithmic binning. Specific bins can also be specified
            with as a string of comma separated durations (e.g "0,1ms,2ms,3ms,4ms,5ms")
            [default: "auto"]

  --format  Format to output histogram in. Supported formats are
            html, json and text. [default: "text"]

Examples:
  echo "GET http://:80" | vegeta attack -name=50qps -rate=50 -duration=5s > results.50qps.bin
  cat results.50qps.bin | vegeta histogram
```

What are your thoughts on this?

fxkr:
I feel that histograms are underused in general, and I would like to see their use promoted.

I would go the opposite way and eventually include a histogram in the default report output. Having it in a separate command kind of implies it's something not normally needed. Whereas in reality averages alone can be misleading, so imo the histogram should always be shown at least for a quick look. As a user, I would also like to see a good enough summary with one command and no options, ie `vegeta report`.

The limit is that I also don't want to have to scroll the output. But the current text output is 7 lines plus errors, so some space is available. And actually I'd rather have the histogram than the errors, which are excessive for me due to #399, but that can be fixed with better aggregation (let's discuss that in the other issue).

I'm not against having subcommands for the individual reports. But:

a) I'd still want a combined `report` for ease of use and to have a standard, easily recognizable output text format (e.g. if I see `iperf` output I know what to look for right away. I do that a _lot_.). It's not mutually exclusive with subcommands, it's enough if it just calls all the individual commands with default params for me. But defaults matter, and good defaults can save a lot of time.

b) I'm fine with having a flexible `report` command with many options for individual "sections" of the report. For histograms, I don't foresee more than `--bins`. That can easily stay in `report`. We shouldn't even have too many parameters. Users with very specific requirements will build their own scripts around the JSON or API anyway.

Also I think it's very important to have a single command that outputs all the data in one JSON. This is kind of important for easy integration. It doesn't have to be `report -format=json`, but it can be.

tsenart:
> I feel that histograms are underused in general, and I would like to see their use promoted.

I echo this sentiment.

> I would go the opposite way and eventually include a histogram in the default report output. Having it in a separate command kind of implies it's something not normally needed. Whereas in reality averages alone can be misleading, so imo the histogram should always be shown at least for a quick look.

It's is [because averages can be misleading](https://bravenewgeek.com/everything-you-know-about-latency-is-wrong/) that the default text report includes percentiles. Do you think that many users fail to interpret percentiles, but would instead succeed with a histogram?

> As a user, I would also like to see a good enough summary with one command and no options, ie vegeta report. The limit is that I also don't want to have to scroll the output. But the current text output is 7 lines plus errors, so some space is available. 

> a) I'd still want a combined report for ease of use and to have a standard, easily recognizable output text format (e.g. if I see iperf output I know what to look for right away. I do that a lot.). It's not mutually exclusive with subcommands, it's enough if it just calls all the individual commands with default params for me. 

How are you envisioning this standard report output? Could you show me an illustrated example?

> But defaults matter, and good defaults can save a lot of time.

Automatic binning, then, seems like the way to go. How would you approach this?

> And actually I'd rather have the histogram than the errors, which are excessive for me due to #399, but that can be fixed with better aggregation (let's discuss that in the other issue).

Good call! These errors can should indeed be better aggregated. Thanks for opening #399.

> b) I'm fine with having a flexible report command with many options for individual "sections" of the report. For histograms, I don't foresee more than --bins. That can easily stay in report. We shouldn't even have too many parameters. Users with very specific requirements will build their own scripts around the JSON or API anyway.

> Also I think it's very important to have a single command that outputs all the data in one JSON. This is kind of important for easy integration. It doesn't have to be report -format=json, but it can be.

Agreed and agreed.

fxkr:
> > I would go the opposite way and eventually include a histogram in the default report output. Having it in a separate command kind of implies it's something not normally needed. Whereas in reality averages alone can be misleading, so imo the histogram should always be shown at least for a quick look.
> 
> It's is [because averages can be misleading](https://bravenewgeek.com/everything-you-know-about-latency-is-wrong/) that the default text report includes percentiles. Do you think that many users fail to interpret percentiles, but would instead succeed with a histogram?

I've read that article before. I was actually wondering why vegeta only outputs the 99th percentile, given that (taking the numbers from there) 18% of users will see worse than the 99 **.9**th percentile. I suppose it's in the region where we need to be careful with introducing errors (e.g. due to GC).

Just to be clear, I wasn't arguing against percentiles, I was just arguing for adding histograms (and keeping the percentiles). I think percentiles and histograms serve different purposes.

Percentiles are an easier to track metric, and are easier to relate to end user experience. From a blackbox "how good is this system?" perspective, percentiles might  even be fine.

Histograms help you _understand_ the system. I know I find it hard to interpret percentiles if I know nothing over the distribution. From a few percentiles I can't even tell a single log-normal distribution from two overlaid ones (e.g. due to a cache that hits in x% cases. [more motivation](http://www.brendangregg.com/FrequencyTrails/modes.html)). So the histogram is useful from a whitebox "how do I tune/fix this system?" perspective.

(It would be different if we could look at the [CDF](https://en.wikipedia.org/wiki/Cumulative_distribution_function), ie "all percentiles as a graph". In terms of information that's equivalent to the histogram. I find them less intuitive than histograms.)




> How are you envisioning this standard report output? Could you show me an illustrated example?

Just all the sections one after another.

I wasn't asking to really change much in the format. What I meant was: if we have a single good output that doesn't rely on customization (via subcommands and parameters, e.g. to enable/disable individual sections) then most vegeta reports (incl. those copy/pasted by other people) will look similar. If we encourage users to parametrize the command, the reader will have to visually "parse" the output again every time.

```
Requests      [total, rate]            10000, 1000.08
Duration      [total, attack, wait]    9.999650299s, 9.999204067s, 446.232µs
Latencies     [mean, 50, 95, 99, max]  508.506µs, 505.419µs, 618.511µs, 733.938µs, 12.052322ms
Bytes In      [total, mean]            5159484, 515.95
Success       [ratio]                  99.99%
Status Codes  [code:count]             0:1  200:9999  

Bucket         #     %       Histogram
[0s,    1ms]   9952  99.52%  ###########################################################################
[1ms,   2ms]   40    0.40%   ##############################
[2ms,   4ms]   6     0.06%   ##############
[4ms,   8ms]   1     0.01%   
[8ms,   +Inf]  1     0.01%   

Error Set:
Get http://localhost:8000: net/http: timeout awaiting response headers
```

> Automatic binning, then, seems like the way to go. How would you approach this?

I have a rough draft in my fork ([branch](https://github.com/fxkr/vegeta/tree/auto-log-bins)/[commit](https://github.com/fxkr/vegeta/commit/6f8063bb65587d2dfb47bfa1cae3b1eab8f31802))

Usage: `vegeta report -type=hist`. It's only extending the `hist` report for now, I didn't touch the text report yet.

The problem I'm running into is vegeta has structs in the public API, which means I have to resort to conditionals instead of polymorphism if I want to have a Histogram that manages its own buckets. If Histogram was an Interface this would not be a problem. How attached are you to API compatibility (relative to clean code)?

I picked base 2 and didn't add a way to customize the base, which is probably ok.

It also includes another bugfix/feature that I should probably split out and propose separately (I'm rescaling the bars of the histogram so that the largest bar fits the available width.)

I was also wondering if linear search is best to find the bucket, but haven't benchmarked this yet. Quite possible binary search is slower since len(buckets) is smallish. Anyway that's also a pre-existing issue, so a fine micro-optimization project for later.

I would do a separate PR for the auto log binning after we merge this one here, but it's still good that we're discussing both here until we have an acceptable CLI.

README.md
```diff
@@ -363,6 +363,8 @@ Options:
   --type    Which report type to generate (text | json | hist[buckets]).
             [default: text]
 
+  --hist    Histogram buckets, e.g.: '[0,1ms,10ms]'
```
tsenart: I would call this flag `--buckets` instead and not make the output of the actual histogram optional. It should always be included in both the JSON and text report, with the `--buckets` value determined automatically from the data set when not specified. This, however, would mean that the algorithm for automatic binning must be appropriate for large data sets (i.e. worse-case O(n)).

report.go
```diff
@@ -85,13 +86,22 @@ func report(files []string, typ, output string, every time.Duration) error {
...
-		if len(typ) < 6 {
-			return fmt.Errorf("bad buckets: '%s'", typ[4:])
-		}
 		var hist vegeta.Histogram
-		if err := hist.Buckets.UnmarshalText([]byte(typ[4:])); err != nil {
+		if histStr == "" { // Old way
+			if len(typ) < 6 {
+				return fmt.Errorf("bad buckets: '%s'", typ[4:])
+			}
+			histStr = typ[4:]
```
tsenart: Let's return an error saying that specifying the buckets via the `--type` flag is deprecated and that one should use the `--buckets` flag instead.

fxkr:
> I like the proposed ideas to extend the default text and JSON report to include a histogram by default. Will you follow up with a PR for that? If we do that, we should deprecate the independent hist report.
Yes, I'll do a follow up PR.

report.go
fxkr: Ok.

README.md
fxkr: Ok, I'll change it to --buckets (won't get to it today, but soon).

I'll do the change to enable the histogram by default using automatic bins in a follow up PR once this one is in.


> This, however, would mean that the algorithm for automatic binning must be appropriate for large data sets (i.e. worse-case O(n)).

Yes. It's already O(num of samples), and num of bucket is "small" (<=16 in practice, <= 64 technically) so that's ok too.

Your comment did make me realize I can just preallocate 64 buckets and remove the extension logic.

I wanted to benchmark the bucket search code, but with 16 buckets I expect linear search will be close to ideal.

fxkr:
Having used the histogram data in the JSON a bit, I now feel that using an array of bucket counts was a huge mistake. I keep second guessing what the bins were in old JSON output. I think I will switch to using a dict. This will also work better with the automatic bins.

fxkr:
- [ ] Change --hist to --buckets
- [ ] In JSON output, change array of bucket counts to map[latency]count
- [ ] Add deprecation warning to --type=hist

tsenart:
> Add deprecation warning to --type=hist

Maybe the deprecation warning should only be added once we have the text report with the histogram included.

ruiaraujo:
Would it be an option to use the HDR histogram log format so that a lot of visualization tooling for HDR can be used?

lib/histogram.go
```diff
@@ -35,6 +36,25 @@ func (h *Histogram) Add(r *Result) {
 	h.Counts[i]++
 }
 
+// MarshalJSON returns a JSON encoding of the buckets and their counts.
+func (h *Histogram) MarshalJSON() ([]byte, error) {
+	var buf bytes.Buffer
+
+	// Custom marshalling to guarantee order.
+	buf.WriteString("{")
+	for i, _ := range h.Buckets {
```
golangcibot: S1005: should omit value from range; this loop is equivalent to `for i := range ...` (from `gosimple`)

fxkr:
> Would it be an option to use the HDR histogram log format so that a lot of visualization tooling for HDR can be used?

I like the idea, but I would say it's out of scope for this PR. I had a brief look at the source code (it appears there's no actual spec for the format) and it would definitely need to be implemented completely separately from vegeta's proprietary JSON format. I don't know what @tsenart thinks about it, but I would suggest filing a separate feature request issue for this.

fxkr:
@tsenart Apologies for the delay! Here's the new changeset. Functionally I'd say it's complete I would do separate PR's for more features (automatic bins, extending text report etc.). Let me know if that works for you or if you find anything wrong with the code. Thanks!

tsenart:
> Would it be an option to use the HDR histogram log format so that a lot of visualization tooling for HDR can be used?

Thanks for bringing this to our attention! I wasn't familiar with it. I think it's a great idea to make Vegeta able to output this.

> d it would definitely need to be implemented completely separately from vegeta's proprietary JSON format.

Veget'a JSON format is not proprietary. Did you mean "custom"? :-)

README.md
```diff
@@ -434,6 +436,7 @@ The `Error Set` shows a unique set of errors returned by all issued requests. Th
     "99th": 3530000,
     "max": 3660505
   },
+  "buckets": {"0":9952,"1000000":40,"2000000":6,"3000000":0,"4000000":0,"5000000":2},
```
tsenart: This seems under-specified from a user documentation perspective. Do you think that making the keys the actual range that the bucket contains would reduce the need for additional documentation?

ruiaraujo:
@tsenart I have a fork that outputs the format I described, I will clean it up and propose a PR later.
The main issue is that the golang implementation of hdr histogram is deprecated with no replacement.

README.md
fxkr: There's two (mutually unrelated) things I see that could be changed, and I don't think we should do either:

1. Use the exact timespec given to vegeta, e.g. "1s" instead of "1000000".
* Makes it harder to parse the output in a unified way. The current code forces a uniform output format, no matter the input. So it's easy to write a script to analyze arbitrary vegeta JSON outputs. Parsing the time specs completely correctly is non-trivial (for non-Go code).

2. Use "${low}-${high}" as key, instead of "${low}".
* The input format wouldn't match the output format.
* Harder for downstream scripts to calculate dict keys.
* You'd need a special case for the overflow bin (e.g. "${low-}infty"). 
* It seems somewhat redundant, esp. when read manually (given the already sorted output).
* Allows corrupt values like "200-100", which is at least a hint of a bad design.
* Downstream scripts will probably generally need to sort the keys (most languages dict's are not ordered). With "${low}" that's easier (map to int, sort), with "${low}-${high}" there's an extra string parsing step.

Documentation-wise, since we print the bins in order, and there's generally a "0" bin, I think it's somewhat obvious what they mean from the example.

Caveat: The docs don't mention that the numbers are nanosec. Somewhat of an oversight. Maybe I should fix this?

Let me know if I should update the code, docs or both. I'm leaning towards only extending the docs with basically "the numbers are nanosecond lower bounds of the bins".

fxkr:
> Veget'a JSON format is not proprietary. Did you mean "custom"? :-)

Yes; that was badly worded. I meant  "non-standard", to contrast with the hdrhistogram format (which is somewhat less non-standard in the way that some other tools probably use it, even though it's still totally not a standard)

README.md
tsenart: Docs update sounds good to me.

tsenart:
@ruiaraujo, @fxkr: Mind to take a look? https://github.com/tsenart/vegeta/pull/413

README.md
fxkr: Done.

fxkr:
Rebased onto master and extended README with description of bucket field in JSON:

```diff
+In the `buckets` field, each key is a nanosecond value representing the lower bound of a bucket.
+The upper bound is implied by the next higher bucket.
+Upper bounds are non-inclusive.
+The highest bucket is the overflow bucket; it has no upper bound.
+The values are counts of how many requests fell into that particular bucket.
```

PR #398, opened 2019-04-28
plot: Sort data points by time elapsed

### LINKED ISSUE #391: Plot not starting at t=0 when there are error responses
#### Version and Runtime

v12.3.0@7bf09f4fab4d852141935796455c17459d212098

#### Expected Behaviour

`vegeta plot` creates a graph at t=0

#### Actual Behaviour

`vegeta plot` created a graph starting at t=892

vegeta-plot (2)
[exp-009-read.html.zip](https://github.com/tsenart/vegeta/files/3095089/exp-009-read.html.zip)

#### Steps to Reproduce

I used vegeta as a library in my load test, against DynamoDB, to test what happens when requests go over the provisioned capacity. My requests started getting throttled at around the 15 minute mark, which is exactly where the plot starts. I'm guessing a bug in plotting specifically.

Unfortunately I can't share the repo, but the setup I think is pretty standard:

```go
func main() {
	rate := vegeta.Rate{Freq: flags.RPS, Per: time.Second}
	duration := flags.Duration
	sess := session.Must(session.NewSession(&aws.Config{
		Region: &flags.AWSRegion,
	}))
	svc := dynamodb.New(sess)
	attacker := vegeta.NewAttacker()

	var res <-chan *vegeta.Result
	switch flags.Op {
	case "write":
		targeter := loadtest.DynamoWriteTargeter(svc, flags.TableName)
		res = attacker.Attack(targeter, rate, duration, "PutItem")
	case "read":
		targeter, err := loadtest.DynamoReadTargeter(svc, flags.TableName, flags.ReadSetSize)
		if err != nil {
			return errors.Wrapf(err, "failed to create read targeter")
		}
		res = attacker.Attack(targeter, rate, duration, "GetItem")
	default:
		logrus.Fatalf("invalid op %s", flags.Op)
	}
	sig := make(chan os.Signal, 1)
	signal.Notify(sig, os.Interrupt)

	var out io.WriteCloser
	out, err := file(flags.Out, true)
	if err != nil {
		return err
	}
	defer out.Close()
	if strings.HasSuffix(flags.Out, ".gz") {
		out = gzip.NewWriter(out)
		defer out.Close()
	}
	enc := vegeta.NewEncoder(out)

encodeLoop:
	for {
		select {
		case <-sig:
			attacker.Stop()
			return nil
		case r, ok := <-res:
			if !ok {
				break encodeLoop
			}
			if err := enc.Encode(r); err != nil {
				return err
			}
		}
	}
}
```

tsenart:
This PR makes the `Plot.data` method return `dataPoints` sorted by
their X axis value (i.e. elapsed seconds)

Previously, each series would be added discontinuously to each other,
essentially not being merge-sorted correctly.

Additionally, the output color palettes were made more intuitive. Reds for
any labels that contain the word ERROR. Greens otherwise.

Any plot with more series than the available colors will be hard to
read because the colors will cycle.

Fixes #391


tsenart:
@mightyguava: Please take a look. This is how your data will look like after this PR. Please not that I set the log-scale to 50 in the HTML plot, but that's not the default.

vegeta-plot

mightyguava:
Looks good. Thank you!

zachbadgett:
IMO this PR makes the plots harder to read in some cases; maybe a similar flag to `-name` should be added called `-color` to override the default or, if that's too specific of a flag, add the ability to add labels, then the plot command can read those.

tsenart:
@zachbadgett: What visual changes do you think would make things easier to read? Can you provide some specific examples?

zachbadgett:
@tsenart it's when you have multiple shades of same color (e.g. two green colors) they can blend in with each other. Being able to set distinguishable colors would help with that. For an example, `user` wants to compare the results from `100qps` and `200qps` ; current functionality would put them on the chart as two different shades of green which can be hard to read. With what I'm suggesting, the `user` can pass which color they would want to use during the attack:
```bash
vegeta attack -name=100qps -label color=green -label error=#EE7860
vegeta attack -name=200qps -label color=blue
```
The labels get passed downstream, `plot` will read the labels it supports and renders the results with the colors specified.

PR #400, opened 2019-04-30
Introduce Pacer interface, ConstantPacer and SinePacer.

fluffle:
#### Background

@tsenart @LasTshaMAN

Supersedes #380, #388, #390.

I wanted to be able to vary attack rates over time, for Reasons. 

The Pacer interface allows vegeta's attack rate to be customized. Attack code keeps track of the elapsed attack duration and the number of hits sent and provides these to Pace, which must return a Duration
for the attack code to sleep for before sending the next hit. Pace may also return true for the second stop value to terminate the attack.

This interface allows vegeta to model a wider range of attack scenarios, such as diurnal request patterns or load-test behaviour, where the attack rate increases until the target begins to serve errors.

The change is primarily structured so that code using the vegeta lib will continue to function as before.

#### Checklist

- [✔] Git commit messages conform to [community standards](http://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html).
- [✔] Each Git commit represents meaningful milestones or atomic units of work.
- [✔] Changed or added code is covered by appropriate tests.

### DIFF
--- a/lib/pacer.go
+++ b/lib/pacer.go
@@ -1,0 +1,64 @@
+package vegeta
+
+import (
+	"fmt"
+	"math"
+	"time"
+)
+
+// A Pacer defines the rate of hits during an Attack by
+// returning the duration an Attacker should wait until
+// hitting the next Target. If the second return value
+// is true, the attack will terminate.
+type Pacer interface {
+	Pace(elapsed time.Duration, hits uint64) (wait time.Duration, stop bool)
+}
+
+// A PacerFunc is a function adapter type that implements
+// the Pacer interface.
+type PacerFunc func(time.Duration, uint64) (time.Duration, bool)
+
+// Pace implements the Pacer interface.
+func (pf PacerFunc) Pace(elapsed time.Duration, hits uint64) (time.Duration, bool) {
+	return pf(elapsed, hits)
+}
+
+// A ConstantPacer defines a constant rate of hits for the target.
+type ConstantPacer struct {
+	Freq int           // Frequency (number of occurrences) per ...
+	Per  time.Duration // Time unit, usually 1s
+}
+
+// Rate is a type alias for ConstantPacer for backwards-compatibility.
+type Rate = ConstantPacer
+
+// ConstantPacer satisfies the Pacer interface.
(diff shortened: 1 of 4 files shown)


lib/pacer_test.go
```diff
@@ -0,0 +1,224 @@
...
+	for ti, tt := range tests {
+		sp := tt.Pacer(0)
+		if got := sp.hits(sp.Period); got != 0 {
+			t.Errorf("%d: %+v.hits(%s) = %g, want 0",
+				ti, sp, sp.Period, got)
+		}
+	}
+}
+
+func TestSinePacerPace_Flat(t *testing.T) {
```
tsenart: What is this test covering exactly?

lib/pacer.go
```diff
@@ -0,0 +1,216 @@
...
+//   H = Mt - (AP/2𝛑)cos(O+(2𝛑/P)t) + (AP/2𝛑)cos(O)
+// Where:
+//   H = Total number of hits triggered during t
+type SinePacer struct {
+	// The period of the sine wave, e.g. 20*time.Minute
+	// MUST BE > 0
+	Period time.Duration
+	// The mid-point of the sine wave in freq-per-Duration,
+	// MUST BE > 0
+	Mean ConstantPacer
```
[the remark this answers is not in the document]
fluffle: OK, fair enough. I went with this approach because I expected the type alias to have a limited lifetime, but if you're wanting to keep it around in perpetuity then using the Rate alias here does make sense.

lib/pacer_test.go
fluffle: It validates that the code does the right thing when there's no amplitude, i.e. the sine wave is actually a straight line. I've added a comment indicating this.

fluffle:
I'd never considered you might want to expose it in the CLI, I've got to admit.

I think you need to delegate parsing the flag args to the right pacer impl. The rate flag could look a bit like "constant:100/1s" or "sine:1h,100/1s,50/1s", which rateFlag would use with a map[string]func(string) (Pacer, error)  by splitting the flag value on ":". Just handwaving here, don't take these suggestions as gospel.

PR #413, opened 2019-06-17
report: HDR Histogram plot output format

tsenart:
Plottable by https://hdrhistogram.github.io/HdrHistogram/plotFiles.html


report.go
```diff
@@ -86,16 +83,21 @@ func report(files []string, typ, output string, every time.Duration) error {
...
+		switch {
+		case strings.HasPrefix(typ, "hist"):
+			if len(typ) < 6 {
+				return fmt.Errorf("bad buckets: '%s'", typ[4:])
+			}
+			var hist vegeta.Histogram
+			if err := hist.Buckets.UnmarshalText([]byte(typ[4:])); err != nil {
+				return err
+			}
+			rep, report = vegeta.NewHistogramReporter(&hist), &hist
```
golangcibot: ineffectual assignment to `rep` (from `ineffassign`)

lib/reporters.go
```diff
@@ -109,3 +109,134 @@ func NewJSONReporter(m *Metrics) Reporter {
...
+	return func(w io.Writer) error {
+		if _, err := w.Write(hdrHeader); err != nil {
+			return err
+		}
+
+		const fmtstr = "%f    %f        %d            %f\n"
+		total := m.Requests
+
+		for _, q := range logarithmic {
+			value := m.Latencies.Quantile(q).Seconds() * 1000 // milliseconds
```
fxkr: This goes through an extra unnecessary float operation (ns -> s -> ms) (potentially introducing errors. might not happen for our number ranges, not sure), and also doesn't use the useful `time` constants. Why not:

```
 int64(m.Latencies.Quantile(q) / time.Millisecond)
```

lib/reporters.go
```diff
@@ -109,3 +109,134 @@ func NewJSONReporter(m *Metrics) Reporter {
...
+
+		return nil
+	}
+}
+
+func oneByQuantile(q float64) float64 {
+	if q < 1.0 {
+		return 1 / (1 - q)
+	}
+	return float64(10000000)
```
fxkr: Where's this constant coming from? (Comment would be helpful)

report.go
```diff
@@ -86,16 +83,21 @@ func report(files []string, typ, output string, every time.Duration) error {
...
-		var hist vegeta.Histogram
-		if err := hist.Buckets.UnmarshalText([]byte(typ[4:])); err != nil {
-			return err
-		}
-		rep, report = vegeta.NewHistogramReporter(&hist), &hist
+	case "hdrplot":
+		var m vegeta.Metrics
+		rep, report = vegeta.NewHDRHistogramPlotReporter(&m), &m
 	default:
+		switch {
```
fxkr: I do not understand why this change is being made as part of this PR?

report.go
```diff
@@ -86,16 +83,21 @@ func report(files []string, typ, output string, every time.Duration) error {
...
+			if len(typ) < 6 {
+				return fmt.Errorf("bad buckets: '%s'", typ[4:])
+			}
+			var hist vegeta.Histogram
+			if err := hist.Buckets.UnmarshalText([]byte(typ[4:])); err != nil {
+				return err
+			}
+			rep, report = vegeta.NewHistogramReporter(&hist), &hist
+		}
 		return fmt.Errorf("unknown report type: %q", typ)
```
fxkr: With this change you'll always return the error for -typ=hist, as you have no return after the `rep, report =` line.

lib/reporters.go
```diff
@@ -109,3 +109,134 @@ func NewJSONReporter(m *Metrics) Reporter {
...
+
+// NewHDRHistogramPlotReporter returns a Reporter that writes out latency metrics
+// in a format plottable by http://hdrhistogram.github.io/HdrHistogram/plotFiles.html.
+func NewHDRHistogramPlotReporter(m *Metrics) Reporter {
+	return func(w io.Writer) error {
+		if _, err := w.Write(hdrHeader); err != nil {
+			return err
+		}
+
+		const fmtstr = "%f    %f        %d            %f\n"
```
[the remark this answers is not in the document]
ruiaraujo: That is the v2 of the format which is more efficient form.

lib/reporters.go
tsenart: We don't want a integer rounding so I used the float conversion method the stdlib uses, but for milliseconds.

tsenart:
@fxkr: Please take another look.

ruiaraujo:
Looking forward to have this released!

@tsenart many thanks for doing this 😃

tsenart:
It’s released!

PR #420, opened 2019-07-19
metrics: Add Throughput to Metrics

tsenart:
This commit adds a `Throughput` field to the `Metrics` struct which
represents the rate of successful requests per second. It is calculated
by dividing the number of successful requests by the time span in
seconds starting from the beginning of the earliest request in the
result set and ending in the time the very last response arrived.

### DIFF
--- a/internal/cmd/echosrv/main.go
+++ b/internal/cmd/echosrv/main.go
@@ -1,15 +1,19 @@
 package main
 
 import (
+	"flag"
 	"log"
 	"net/http"
 	"net/http/httputil"
-	"os"
 	"sync/atomic"
 	"time"
 )
 
 func main() {
+	sleep := flag.Duration("sleep", 0, "Time to sleep per request")
+
+	flag.Parse()
+
 	count := uint64(0)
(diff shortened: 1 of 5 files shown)


internal/cmd/echosrv/main.go
```diff
@@ -20,8 +24,10 @@ func main() {
 		}
 	}(time.Now())
 
-	http.ListenAndServe(os.Args[1], http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		atomic.AddUint64(&count, 1)
+	http.ListenAndServe(flag.Arg(0), http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
golangcibot: Error return value of `http.ListenAndServe` is not checked (from `errcheck`)

raul1991:
@tsenart  Superb tool! 

Upon reading the documentation, I couldn't find a way to plot the throughput captured as part of the report. Can the doc be updated with that or is there any link I can read it on ?

PR #421, opened 2019-07-19
attack: Add -max-workers flag

tsenart:
This PR introduces the `-max-workers` flag and associated function
options. This allows controlling the concurrency level that vegeta uses,
but can sacrifice reaching the requested `-rate`, in case the specified
max number of workers is below the what is needed to sustain such rate.

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,85 +51,87 @@ Usage: vegeta [global flags] <command> [command flags]
 
 global flags:
   -cpus int
-    	Number of CPUs to use (defaults to the number of CPUs you have)
+        Number of CPUs to use (default 16)
   -profile string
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -51,85 +51,87 @@ Usage: vegeta [global flags] <command> [command flags]
 
 global flags:
   -cpus int
-    	Number of CPUs to use (defaults to the number of CPUs you have)
+        Number of CPUs to use (default 16)
```
tsenart:
```suggestion
        Number of CPUs to use (defaults to the number of CPUs you have)
```

tsenart:
@keegancsmith, @mrnugget: PTAL

PR #423, opened 2019-07-20
attack: Benchmarking with -rate=0 and -max-workers=N

### LINKED ISSUE #378: Proposal: Make rate limit optional
#### Proposal

It seems like Vegeta requires a request rate. It must either be specified with `-rate` or a default rate of 1 per worker per second is used. I propose making this optional, leading to each worker making another request as soon as the previous one finished.
This is related to, but in my opinion different from https://github.com/tsenart/vegeta/issues/163 and https://github.com/tsenart/vegeta/issues/271.

#### Background

I'd like to know how many requests per second my web service supports at a reasonable response time.

Ramping up the rate (e.g. attacking for 5s with 50 requests second, then for 5 seconds with 100 requests per second etc.) makes it possible to "guess" the maximum request rate, because at some point the response times suddenly take much longer than before, but this is a little cumbersome to analyze to get precise numbers.

To expand on the previous paragraph regarding what I mean with "guessing" the maximum request rate: Let's say a rate of 50 leads to 30ms avg response time, a rate of 100 leads to 35ms avg response time, a rate of 200 leads to 50ms avg response time, then suddenly a rate of 400 leads to 5s response time. I would guess that the web service supports around 300 requests per second (overall, with 50 clients). To get a more precise number the increase of the rate can now be lowered and another round of load testing with rates between 200 and 400 requests is required.

I think this number can be determined easier and more precisely with the proposed approach.

#### Workarounds

Other load testing tools seem to be focused on exactly this approach, one of them being https://github.com/rakyll/hey.
Example: `hey -z 10s http://localhost:8080`
The first part of the result is:
```
Summary:
  Total:        10.0065 secs
  Slowest:      0.0237 secs
  Fastest:      0.0003 secs
  Average:      0.0065 secs
  Requests/sec: 7646.6975
...(snip)
```
Where `Requests/sec` is the number I'm interested in.

But I think in my specific case (a web service with large response bodies) the results I get from `hey` are off and Vegeta also seems to be more actively maintained, which is why I would like to see this feature in Vegeta.

tsenart:
This commit allows setting a zero rate which, together with a reasonable
number of `-max-workers`, behaves as independent concurrent users, each
sending a request and waiting for the response before sending the next
one.

This is part of enabling vegeta to work with the Universal Scalability
Law model, where a concurrency unit is defined as an independent
thread/worker/driver as described above (i.e. with backpressure).

Fixes #378

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,87 +51,87 @@ Usage: vegeta [global flags] <command> [command flags]
 
 global flags:
   -cpus int
-        Number of CPUs to use (defaults to the number of CPUs you have)
+    	Number of CPUs to use (defaults to the number of CPUs you have)
   -profile string
-        Enable profiling of [cpu, heap]
+    	Enable profiling of [cpu, heap]
   -version
-        Print version and exit
+    	Print version and exit
 
 attack command:
   -body string
-        Requests body file
+    	Requests body file
   -cert string
-        TLS client PEM encoded certificate file
+    	TLS client PEM encoded certificate file
   -connections int
-        Max open idle connections per target host (default 10000)
+    	Max open idle connections per target host (default 10000)
   -duration duration
-        Duration of the test [0 = forever]
+    	Duration of the test [0 = forever]
   -format string
-        Targets format [http, json] (default "http")
+    	Targets format [http, json] (default "http")
   -h2c
-        Send HTTP/2 requests without TLS encryption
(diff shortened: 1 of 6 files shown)


tsenart:
@mrnugget, @keegancsmith: PTAL

tsenart:
@mrnugget: Implemented your suggestion and added some documentation.

Issue #418, opened 2019-07-15
Add the ability to ramp-up the load 


domleb:
#### Proposal

Add the ability to ramp-up the load.

#### Background

I see the has already been requested [here](https://github.com/tsenart/vegeta/issues/163 ) and closed in favor of invoking the command multiple times with different attack rates. 

The trouble with this approach is that it creates connection spikes every time the load changes because the existing connections are not reused.  This results in larger spikes as the load increases.  At a certain point this will cause connection errors as most target systems have a limit on the number of instantaneous connections they can create.  For example, this will cause connection errors with ELBs because the connection rate does not match the request rate for a given ELB instance.  This is the problem I have.

#### Workarounds

This problem can be avoided by running multiple vegeta instances at the same time and staggering the start time.

I would be happy to implement this if we can get a design together.

tsenart:
We already have all the machinery for this in here: https://github.com/tsenart/vegeta/blob/master/lib/pacer.go#L118

How would you see this exposed via the CLI?

domleb:
I did see the pacer change.  It's an unusual feature (for a load testing tool) and there isn't much documentation yet so I wasn't sure how to use it for a ramp-up.  It's worth being careful with naming too, as historically tools use the term 'pacing' to control iterations and 'ramp-up' to ramp-up to a given pace.

Here's the DSL for gatling:
```
setUp(
  scn.inject(
    nothingFor(4 seconds), // 1
    atOnceUsers(10), // 2
    rampUsers(10) over (5 seconds), // 3
    constantUsersPerSec(20) during (15 seconds), // 4
    constantUsersPerSec(20) during (15 seconds) randomized, // 5
    rampUsersPerSec(10) to 20 during (10 minutes), // 6
    rampUsersPerSec(10) to 20 during (10 minutes) randomized, // 7
    splitUsers(1000) into (rampUsers(10) over (10 seconds)) separatedBy (10 seconds), // 8
    splitUsers(1000) into (rampUsers(10) over (10 seconds)) separatedBy atOnceUsers(30), // 9
    heavisideUsers(1000) over (20 seconds) // 10
  ).protocols(httpConf)
)
```

This is obviously a very rich API.  The simplest CLI argument would be `-ramp 60s` like the simple ramp-up in earlier versions of LoadRunner.

Or something more elaborate like the suggestions in https://github.com/tsenart/vegeta/issues/163. Here's another example that could also be passed in as a csv file:

[...]

PR #425, opened 2019-07-23
pacer: Introduce LinearPacer

tsenart:
This commit introduces a LinearPacer which paces an attack by starting
at a given request rate and increasing linearly with the given slope.

This is a pre-requisite for implementing #418

### DIFF
--- a/lib/pacer.go
+++ b/lib/pacer.go
@@ -217,3 +217,63 @@ func (sp SinePacer) hits(t time.Duration) float64 {
 	}
 	return sp.Mean.hitsPerNs()*float64(t) + sp.ampHits()*(math.Cos(sp.StartAt)-math.Cos(sp.radians(t)))
 }
+
+// LinearPacer paces an attack by starting at a given request rate
+// and increasing linearly with the given slope.
+type LinearPacer struct {
+	StartAt Rate
+	Slope   float64
+}
+
+// Pace determines the length of time to sleep until the next hit is sent.
+func (p LinearPacer) Pace(elapsed time.Duration, hits uint64) (time.Duration, bool) {
+	switch {
+	case p.StartAt.Per == 0 || p.StartAt.Freq == 0:
+		return 0, false // Zero value = infinite rate
+	case p.StartAt.Per < 0 || p.StartAt.Freq < 0:
+		return 0, true
+	}
+
+	expectedHits := p.hits(elapsed)
+	if hits == 0 || hits < uint64(expectedHits) {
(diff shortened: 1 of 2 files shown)


tsenart:
PTAL @mrnugget, @keegancsmith

lib/pacer.go
```diff
@@ -217,3 +217,63 @@ func (sp SinePacer) hits(t time.Duration) float64 {
...
+	if hits == 0 || hits < uint64(expectedHits) {
+		// Running behind, send next hit immediately.
+		return 0, false
+	}
+
+	rate := p.rate(elapsed)
+	interval := math.Round(1e9 / rate)
+
+	if n := uint64(interval); n != 0 && math.MaxInt64/n < hits {
+		// We would overflow delta if we continued, so stop the attack.
```
tsenart:
```suggestion
		// We would overflow wait if we continued, so stop the attack.
```

tsenart:
cc @fluffle

lib/pacer.go
```diff
@@ -217,3 +217,63 @@ func (sp SinePacer) hits(t time.Duration) float64 {
...
+func (p LinearPacer) Pace(elapsed time.Duration, hits uint64) (time.Duration, bool) {
+	switch {
+	case p.StartAt.Per == 0 || p.StartAt.Freq == 0:
+		return 0, false // Zero value = infinite rate
+	case p.StartAt.Per < 0 || p.StartAt.Freq < 0:
+		return 0, true
+	}
+
+	expectedHits := p.hits(elapsed)
+	if hits == 0 || hits < uint64(expectedHits) {
```
keegancsmith: why do you need to compare hits to 0? I would expect expectedHits to always be > 0

tsenart: It'd be zero if elapsed == 0. That shouldn't happen at runtime in the vegeta attack command, only in tests, or when called as a library with that value.

anas-qa:
@tsenart i was about to open a question issue for ramp-up and ramp-down but then i saw this.

Readme.md is not up to date with `pacer`

Should i expect a doc soon

tsenart:
@anas-qa: I haven't had the time to implementing #418, so there's no way to use this via the CLI yet.

anas-qa:
Thanks @tsenart 

This would be a huge +1 once it is

Hope soon 👍

fluffle:
Sorry I didn't get around to this, life sort-of got in the way. The code looks great, this is exactly the sort of thing you need to do destructive load-testing :-)

Issue #416, opened 2019-07-01
Support HTTP CONNECT proxy headers


SalmaanP:
#### Question

I see there is a way to make plain HTTP requests, just wondering if its possible to make HTTP CONNECT requests? So the client would specify HTTP CONNECT in the plain request with some extra headers optionally, the proxy would reply with `HTTP 200 OK Connection Established`. SSL handshake starts with the client sending client hello. If someone has tried this already, is there an example? Thanks!

tsenart:
Have you tried it?

SalmaanP:
Yes, I tried it out with setting the `HTTPS_PROXY` environment variable set, and it works as expected. The only thing I can figure out is how to send CONNECT headers. In curl you can specify it using `--proxy-headers`. I can see that the CONNECT request already has `User-Agent: Go-http-client/1.1` added so it might be just adding a conf?

tsenart:
Gotcha. It seems like we'd have to implement another `AttackerOpt` that sets `ProxyConnectHeader` in https://golang.org/pkg/net/http/#Transport, and expose that as `vegeta attack` flag (e.g. `-proxy-header`).

SalmaanP:
I'm gonna try and pick this up this week.

PR #434, opened 2019-09-04
Support HTTP CONNECT proxy headers

SalmaanP:
#### Background

This PR adds support for proxy CONNECT headers. This allows sending custom headers to the proxy using the `-proxy-header` flag. Multiple headers can be sent by reusing the flag. Fixes https://github.com/tsenart/vegeta/issues/416

Example:
```
export HTTPS_PROXY=http://localhost:12345
./vegeta attack -duration=1s -insecure -targets=targets.txt -proxy-headers "Header1:Key1" -proxy-headers "Header2:key2"
[tmp] nc -lv 12345
CONNECT www.example.com:443 HTTP/1.1
Host: www.example.com:443
User-Agent: Go-http-client/1.1
Header1: Key1
Header2: key2

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,8 +92,8 @@ attack command:
     	Attack name
   -output string
     	Output file (default "stdout")
-  -proxy-headers value
-    	Headers for the CONNECT request
+  -proxy-header value
+    	Header for the CONNECT request
   -rate value
     	Number of requests per time unit [0 = infinity] (default 50/1s)
   -redirects int
--- a/README.md
+++ b/README.md
@@ -93,7 +93,7 @@ attack command:
   -output string
     	Output file (default "stdout")
   -proxy-header value
-    	Header for the CONNECT request
+    	Proxy CONNECT header
   -rate value
     	Number of requests per time unit [0 = infinity] (default 50/1s)
   -redirects int
--- a/README.md
+++ b/README.md
@@ -92,6 +92,8 @@ attack command:
     	Attack name
   -output string
     	Output file (default "stdout")
+  -proxy-headers value
+    	Headers for the CONNECT request
   -rate value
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: README.md [hunk 1]
```diff
@@ -92,6 +92,8 @@ attack command:
     	Attack name
   -output string
     	Output file (default "stdout")
+  -proxy-headers value
```
tsenart: This should be singular like the `-header` flag: `-proxy-header`

attack.go
```diff
@@ -133,6 +136,7 @@ func attack(opts *attackOpts) (err error) {
 		tr  vegeta.Targeter
 		src = files[opts.targetsf]
 		hdr = opts.headers.Header
+		proxyHdr = opts.proxyHeaders.Header
```
tsenart: It seems like you haven't ran gofmt on the code. Please do.

lib/attack.go
```diff
@@ -229,6 +229,15 @@ func Client(c *http.Client) func(*Attacker) {
 	return func(a *Attacker) { a.client = *c }
 }
 
+// ProxyHdr returns a functional option that allows you to add your own
+// Proxy CONNECT headers
+func ProxyHdr(h *http.Header) func(*Attacker) {
```
tsenart: Please call this `ProxyHeader`

tsenart: You can pass a non-pointer value since the `http.Header` type is a reference type: map.

lib/attack.go
```diff
@@ -229,6 +229,15 @@ func Client(c *http.Client) func(*Attacker) {
...
+	return func(a *Attacker) {
+		tr := a.client.Transport.(*http.Transport)
```
```suggestion
		if tr, ok := a.client.Transport.(*http.Transport); ok {
		    ...
		}
```

PR #441, opened 2019-09-23
Add header in result

qneyrat:
I added the headers of a Target that's been hit in the Result on vegeta capture.

### DIFF
--- a/lib/results.go
+++ b/lib/results.go
@@ -6,7 +6,9 @@ import (
 	"encoding/base64"
 	"encoding/csv"
 	"encoding/gob"
+	"encoding/json"
 	"io"
+	"net/http"
 	"sort"
 	"strconv"
 	"time"
@@ -32,6 +34,7 @@ type Result struct {
 	Body      []byte        `json:"body"`
 	Method    string        `json:"method"`
 	URL       string        `json:"url"`
+	Header    http.Header   `json:"header"`
 }
 
 // End returns the time at which a Result ended.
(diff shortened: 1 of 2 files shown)


lib/results.go
```diff
@@ -156,6 +173,7 @@ func NewCSVEncoder(w io.Writer) Encoder {
 			strconv.FormatUint(r.Seq, 10),
 			r.Method,
 			r.URL,
+			string(header),
```
tsenart: This would break CSV parsing. We'd need to encode this differently. Perhaps Base64 encoding?

lib/results.go
```diff
@@ -32,13 +34,22 @@ type Result struct {
...
 }
 
 // End returns the time at which a Result ended.
 func (r *Result) End() time.Time { return r.Timestamp.Add(r.Latency) }
 
 // Equal returns true if the given Result is equal to the receiver.
 func (r Result) Equal(other Result) bool {
+	headerIsEqual := true
+	for headerKey := range r.Header {
+		if r.Header.Get(headerKey) != other.Header.Get(headerKey) {
```
tsenart: This only gets the last value for a given Header key. Fix that and extract this code into a function we  can call below in the return statement.

```go
func headersEqual(a, b http.Header) bool {
    for k, as := range a {
        bs := b[k]
        if len(bs) != len(as) {
            return false
        }
        for i := range as {
            if as[i] != bs[i] {
                return false
            }
        }
    }
    return true
}
```

tsenart:
Fixed by #474

PR #446, opened 2019-10-07
Report minimum latency

0Delta:
#### Background

I tried create box-and-whisker plot about latency using `vegeta report` .
Be in trouble vegeta did not tell me minimum latency, box-and-whisker plot need max,3/4,1/2,1/4 and minimum value.
As an alternative i used max,50th and 95th value. but i can't replaced min value, i thought replace min value for someone is not useful.

#### Checklist


lib/metrics.go
```diff
@@ -157,6 +163,7 @@ func (l LatencyMetrics) Quantile(nth float64) time.Duration {
 
 func (l *LatencyMetrics) init() {
 	if l.estimator == nil {
+		l.Min = time.Duration(math.MaxInt64)
```
tsenart: What makes this line necessary?

0Delta: Because `Int` value is initialize with `0`  .
But, I forgot measures were already taken at [Line 152].(https://github.com/tsenart/vegeta/pull/446/files/3fa39a80ed8dc214606ea630304620fb65a84db2#diff-d60ed9ab81a6041797ea9d191d7bb753R152)

I remove it. thank you for check.

0Delta:
Sorry for late response.
I fix about unused code and add document about the new value.

Please check and let me know if something is wrong ?

PR #451, opened 2019-10-15
Add web server benchmarking script

fxkr:
#### Background

This script will automatically run vegeta against a target with different request rates and graph the latency distribution and success rate at each request rate.

This lets you see a couple interesting things that are less visible otherwise, e.g.:

* Multimodal latency distributions and outliers
* How performance degrades well _before_ the server starts dropping packets

The script is intentionally short (no cli options or abstraction layers etc) and excessively documented, so that people can easily adapt it to their needs.

I'm not sure if doing this as a separate script is ok for you. The original version of this was written in Go, using vegeta's API instead of JSON interface, with config as constants. But that wasn't very useable, especially without an already setup dev environment. I thought about integrating it into vegeta itself, but it'd have been a lot more code, and I couldn't come up with a good CLI.  So I rewrote it as an external script. I found this script useful, and so I thought I'd clean it up and see if you want it. If nothing else it could serve as an example of how to use vegeta's JSON output.

The README and commit message are a bit minimalist. That's cause I didn't want to commit binary files (screenshots). Instead they link to a currently non-existing wiki page - https://github.com/tsenart/vegeta/wiki/Benchmarking -, which I'll write once I know if this PR has any chance of getting merged.

Of course I'd be happy to make whatever changes you'd like. If you'd rather not have such external scripts in the git repo that's fine too.

#### How to use

Dependencies:

* Python 3
* Gnuplot

Usage:

```
echo GET http://localhost:8080/ | ./benchmark.py
```

#### Examples

Here's an example result from `python3 -m http.server` in an empty directory:

image

You can see that there are unacceptable (>1 sec) latency spikes well before it starts dropping requests.

For comparison, [lwan](https://lwan.ws/) serving the same:

image

Unsurprisingly much better. But it's interesting how latency _improves_ when the request rate goes up :-)

#### Checklist

### DIFF
--- a/scripts/load-ramping/README.md
+++ b/scripts/load-ramping/README.md
@@ -1,0 +1,17 @@
+# Load ramping
+
+This script will automatically run vegeta against a target with different request
+rates and graph the latency distribution and success rate at each request rate.
+
+Usage:
+
+```
+echo GET http://localhost:8080/ | python3 ramp-requests.py
+```
+
+Dependencies:
+
+* Python 3
+* Gnuplot
+
+For more documentation, see https://github.com/tsenart/vegeta/wiki/Load-ramping
--- a/scripts/load-ramping/ramp-requests.plt
+++ b/scripts/load-ramping/ramp-requests.plt
@@ -1,0 +1,75 @@
+# Two plots: success rate plot on top, rate/latency distribution below
+set multiplot layout 2,1
+
+
+#
+# Shared config
+#
+
+# Scale (x/color)
+set autoscale xfix
+set logscale xycb 10
+
+# Grid
+set mxtics 10
+set mytics 10
+set tics scale 0.0000000001  # Tics themselves can't be styled indepedently, so use grid only
+set grid xtics ytics mxtics mytics lc rgb '#888888' lw 0.5 lt 1, lc rgb '#888888' lt 1 lw 0.1
+
+
+#
+# Top plot: success rate
+#
+
+# Manual positioning to align both plots
+set lmargin at screen 0.10
+set rmargin at screen 0.87
+set bmargin at screen 0.80
+set tmargin at screen 0.95
+
+# Scale (y only)
+set yrange [1.0:100.0]
+
+# Axes
+unset xlabel
+set xtics format ""
+set ylabel "Success"
+set ytics format "%.2f%%"
+
+# Plot (incl. fraction to percentage conversion)
+plot "results_success.txt" using 1:($2 * 100.0):(0.0) with line lw 3 lc rgb "red" title ""
+
+
+#
+# Bottom plot: rate vs latency
+#
+
+# Manual positioning to align both plots
+set lmargin at screen 0.10
+set rmargin at screen 0.87
+set tmargin at screen 0.75
+set bmargin at screen 0.15
+
+# Scale (y only)
+unset yrange
+set autoscale yfix
+
+# Axes
+set xlabel "Requests (per sec)"
+set xtics format "%.f"
+set ylabel "Latency" offset -1.5,0,0
+set ytics ( \
+    "1ns" 1.0e0, "10ns" 1.0e1, "100ns" 1.0e2, \
+    "1us" 1.0e3, "10us" 1.0e4, "100us" 1.0e5, \
+    "1ms" 1.0e6, "10ms" 1.0e7, "100ms" 1.0e8, \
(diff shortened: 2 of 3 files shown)


tsenart:
This visualisation is awesome. Thank you! A few thoughts:

1. Can you think of a less generic / overloaded term than "benchmarking"?
2. How much more code would it be to make this a vegeta sub-command (or just a `report` type)?

In any case, I'm happy to merge after point number 1 above is discussed.

fxkr:
>  1. Can you think of a less generic / overloaded term than "benchmarking"?

I have no good idea.

So I went to Brendan Gregg's "Systems Performance" book for inspiration, and found chapter "12.3.7. **Ramping Load**", which matches exactly what this script does:

> "This is a simple method for determining the maximum throughput a system can handle. In involves adding load in small increments and measuring the delivered throughput until a limit is reached. The results can be graphed, showing a **scalability profile**. This profile can be studied visually or by using scalability models.
> 
> [...] When following this approach, measure latency as well as the throughput, especially the latency distribution. [...] If you push load too high, latency may become so high that it is no longer reasonable to consider the result as valid. Ask yourself if the delivered latency would be acceptable to a customer."

So maybe "load ramping" or "scalability profiling"? I'm not happy with either of these but they're the best I've got, and fwiw I agree "benchmarking" is even worse. Would either of these work for you?

>  2. How much more code would it be to make this a vegeta sub-command

For a 1:1 feature parity version, not much at all.

But having a script lets us avoid inventing a rather complex CLI. There are a lot of things a user may want to tweak...

* Orchestrating the `attack`s -> target format, same targets for all attacks vs keep reading from stdin as needed, which rates to use, where to store the output, cooldown period between tests
* The individual `attack`s --> any extra flags like -http2
* Extracting the data in gnuplot format --> it's pretty specific to the graph, so if the gnuplot code is separate this should be too
* The definition of the graph --> many many twiddly things
* Rendering the graph --> output formats, filename (I assume we'd stick to an external gnuplot binary?), size

We'd have to make some tradeoff between flexibility and ease of use and I don't think there's any good tradeoff. At least not with the scope that vegeta has currently (it's more like iperf than IxNetwork).

> (or just a `report` type)?

That can't be enough, half of the script is about running the attacks. Sure we could have a `report` type that takes a bunch of files and merges them into a graph, but it increases the chance that it'll be used wrong (e.g. user accidentally combines files generated with different attack options).

tsenart:
I think the load ramping terminology makes sense so let's rename things to that and then I can merge this :-)

lpereira:
This looks really nice!  Wish I had this tool while doing some early work in Lwan.

May I suggest adding another version of this plot, where the X axis is the number of concurrent clients, and the success plot is replaced by (or another series is added in the same plot) the throughput measurement in RPS?  This would help identify scalability issues as the number of clients increase.

fxkr:
@tsenart thanks. I'll get to it over the next couple of days.

@lpereira oh wow. Thanks for lwan! Sure, I can do that. I'll do a separate PR, but here's a preview. Is this about right, and is it useful?

image

The main question I have is how would we scale up the -workers and -max-workers (relative to the -connections)? Currently it's 1:1:1.

(Don't look too closely on the x<10 part. There's some rounding error going on that I haven't dealt with yet.)

(I wonder if lwan outperforms vegeta here, and how we would find that out... goroutines aren't exactly free.)

fxkr:
Renamed to scripts/load-ramping/ramp-requests.py and update commit msg + readme. I chose this naming scheme for the script with the expectation of adding a ramp-conns.py (the script @lpereira requested) in the same directory later.

As promised, I also wrote the missing wiki page: https://github.com/fxkr/vegeta/wiki/Load-ramping

Would it be ok for you to enable public write access to the vegeta wiki so that I can copy it over to your repo? Alternatively I can refine it a bit more and let you know so you can copy it over yourself.

tsenart:
> Would it be ok for you to enable public write access to the vegeta wiki so that I can copy it over to your repo? Alternatively I can refine it a bit more and let you know so you can copy it over yourself.

I'd prefer to keep the Wiki secured for now so let know if you want to take a last look at the document before I copy it over.

fxkr:
Ok. Let's copy it as is for now, I suppose if I get around to making significant changes (e.g. to describe the "connections vs throughput" script) I can ask you again.

fxkr:
(fyi, Github doesn't have UI for adding images to the wiki. The images are added by cloning the wiki's git repo and adding them manually.)

PR #453, opened 2019-10-22
feat(http3): add initial support

VoyTechnology:
Add initial support for HTTP3/QUIC. This might or might not work based on the
server in used and should be treated as experimental until the standard stops
being an IETF Draft.

The PR brings it just to the initial state where you can call a server with `--http3` flag, but it might or might not work based on the supported version of QUIC on the server.

As part of the change I also had to check for the type of the `Transport` used, thus removing a possible panic. While it could have been created as a separate PR it is an integral part of this change. Also, most of the changes in `go.mod` are mostly the result of running `go mod tidy`, which was not ran before.

Testing was not fully complete for this PR, but I welcome the discussion on the best course of action on how to do so.

#### Background

QUIC is the underlying communication in uncoming HTTP3. This PR adds the basic functionality based on the latest version of the most common open source quic implementations in Go: https://github.com/lucas-clemente/quic-go

#### Checklist


lib/attack.go
```diff
@@ -192,15 +202,23 @@ func HTTP2(enabled bool) func(*Attacker) {
...
 // performed by an Attacker
 func H2C(enabled bool) func(*Attacker) {
 	return func(a *Attacker) {
-		if tr := a.client.Transport.(*http.Transport); enabled {
+		if tr, ok := a.client.Transport.(*http.Transport); ok && enabled {
 			a.client.Transport = &http2.Transport{
 				AllowHTTP: true,
 				DialTLS: func(network, addr string, cfg *tls.Config) (net.Conn, error) {
-					return tr.Dial(network, addr)
+					return tr.DialContext(context.TODO(), network, addr)
```
```suggestion
					return tr.DialContext(context.Background(), network, addr)
```

attack.go
```diff
@@ -178,6 +180,7 @@ func attack(opts *attackOpts) (err error) {
 		vegeta.KeepAlive(opts.keepalive),
 		vegeta.Connections(opts.connections),
 		vegeta.HTTP2(opts.http2),
+		vegeta.HTTP3(opts.http3),
```
tsenart: I believe this option needs to come before `TLSConfig` as to not override the side effects that opt has.

lib/attack.go
```diff
@@ -192,15 +202,23 @@ func HTTP2(enabled bool) func(*Attacker) {
 	}
 }
 
+func HTTP3(enabled bool) func(*Attacker) {
+	return func(a *Attacker) {
+		if enabled {
+			a.client.Transport = &http3.RoundTripper{}
```
tsenart: Any chance we can "translate" already set configuration in an existing Transport to this `http3.RoundTripper`? Akin to what `http2.ConfigureTransport(tr)` does. Example: Not loosing existing `TLSClientConfig`.

PR #455, opened 2019-10-27
attack: Implement the -chunked flag

### LINKED ISSUE #439: How to send chunk data by using vegeta?
Does vegeta sends chunk data when Transfer-Encoding header is set to "chunked" while sending data?

tsenart:
Some users of vegeta must be able to control how the request body is
sent. By default, the Go HTTP client doesn't make any guarantees about
which transfer encoding is used.

https://sourcegraph.com/github.com/golang/go/-/blob/src/net/http/request.go#L202

Fixes #439

### DIFF
--- a/internal/cmd/echosrv/main.go
+++ b/internal/cmd/echosrv/main.go
@@ -10,11 +10,13 @@ import (
 	"log"
 	"net/http"
 	"net/http/httputil"
+	"os"
 	"sync/atomic"
 	"time"
 )
 
 func main() {
+	dump := flag.Bool("dump", false, "Dump HTTP requests to stdout")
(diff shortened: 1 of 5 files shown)


internal/cmd/echosrv/main.go
```diff
@@ -42,7 +44,13 @@ func main() {
...
 
 		bs, _ := httputil.DumpRequest(r, true)
-		w.Write(bs)
+
+		out := io.Writer(w)
+		if *dump {
+			out = io.MultiWriter(w, os.Stdout)
+		}
+
+		out.Write(bs)
```
golangcibot: Error return value of `out.Write` is not checked (from `errcheck`)

Issue #462, opened 2019-11-26
Proposal: Add a correlation ID to the Target that is shared with the Result to help trace slow queries.


jacquesbernier:
#### Proposal

**Option 1**:

The proposal is to attach the `Result.Seq` (https://github.com/tsenart/vegeta/blob/master/lib/attack.go#L338-L342) to the `Target` that is passed to the `Targeter` https://github.com/tsenart/vegeta/blob/master/lib/attack.go#L351.
`Result.Seq` is set before the Targeter is called. Therefore, adding the a field `Target.Seq` and set it to the same value is a possibility.
I believe this one can be backward compatible and a low cost option.

**Option 2**:

Add a new option to the `Attacker` that allows the user to provide a correlation ID generator (this can be a function that returns a string). Add the generated correlation ID to the target and the Result in the `Attacker.attack` method.

#### Background

We use vegeta for load testing and we report `Result`s for slow queries. We would love to have a way to correlate a `Result` to the request that was made to help us analyze if there is a systematic problem with particular queries that we use.

We use vegeta in library mode and are able to store the requests. We would want to have a correlation id and we believe that the `Result.Seq` could be it.

Note:
I recognized that it may seem related to https://github.com/tsenart/vegeta/issues/205 to which you answer that server logs could be used. However, I see 2 cases where this may be hard. Logging all requests on the server may affect the performances that we trying to test. In other cases, we may not have that possibility. I.E. we are sometimes load testing against vendors and have little visibility into their logs.

#### Workarounds

By providing a custom http client that intercepts the request and response.
However, this means that the body may be read multiple times.

tsenart:
Would adding an `X-Vegeta-Seq` header to each request that is done be sufficient?

jackdbernier:
The goal is to have the information in the `Result` https://github.com/tsenart/vegeta/blob/44a49c878dd6f28f04b9b5ce5751490b0dce1e18/lib/results.go#L23. The headers are not attached to the Result. That is why I proposed different ways.

tsenart:
@jackdbernier: However, the `Seq` number is already in the `Result`, so if we add a `X-Vegeta-Seq` header to each `Target`, you'd know how to correlate.

jackdbernier:
@tsenart yes. I feel like an idiot now 😄. I think this would be an elegant solution and 100% backward compatible.

tsenart:
[...]

PR #470, opened 2020-01-19
Add header to target containing the sequence number

jackdbernier:
#### Background

Following up the discussion here https://github.com/tsenart/vegeta/issues/462

This adds the res.Seq to the target headers with the name "X-Vegeta-Seq".
This also add a new unit test to check that the header is equal to the res.Seq.

#### Checklist

### DIFF
--- a/lib/attack.go
+++ b/lib/attack.go
@@ -13,6 +13,8 @@ import (
 	"sync"
 	"time"
 
+	"strconv"
+
 	"golang.org/x/net/http2"
 )
 
@@ -331,7 +333,7 @@ func (a *Attacker) attack(tr Targeter, name string, workers *sync.WaitGroup, tic
 func (a *Attacker) hit(tr Targeter, name string) *Result {
 	var (
 		res = Result{Attack: name}
-		tgt Target
+		tgt = Target{Header: make(http.Header)}
 		err error
 	)
 
@@ -347,6 +349,8 @@ func (a *Attacker) hit(tr Targeter, name string) *Result {
 			res.Error = err.Error()
 		}
 	}()
+
+	tgt.Header.Set("X-Vegeta-Seq", strconv.FormatUint(res.Seq, 10))
 
 	if err = tr(&tgt); err != nil {
 		a.Stop()
--- a/lib/attack_test.go
+++ b/lib/attack_test.go
@@ -13,6 +13,7 @@ import (
 	"os"
 	"path/filepath"
 	"reflect"
+	"strconv"
 	"strings"
 	"testing"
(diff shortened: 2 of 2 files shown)


lib/attack_test.go
```diff
@@ -358,3 +359,19 @@ func TestClient(t *testing.T) {
...
+
+	atk := NewAttacker()
+	res := atk.hit(func(trt *Target) error {
+		got = trt.Header
+		return nil
+	}, "")
+
+	if got.Get("X-Vegeta-Seq") != strconv.FormatUint(res.Seq, 10) {
+		t.Fatalf("got: %v, want: %v", got.Get("X-Vegeta-Seq"), res.Seq)
+	}
```
jackdbernier: I'm wondering if it would be better to have a small loop here to make sure that we are not only getting the zero value. Here `res.Seq` is equal to 0.

lib/attack_test.go
```diff
@@ -358,3 +359,21 @@ func TestClient(t *testing.T) {
...
 	}
 }
+
+func TestSeqHeader(t *testing.T) {
+	t.Parallel()
+
+	var got http.Header
+
+	atk := NewAttacker()
+	for i := 0; i < 5; i++ {
```
jackdbernier: Here I'm looping to ensure that we are not only testing `res.Seq == 0`. I can remove it if you prefer.

tsenart: Sounds good, just add that as code comment.

lib/attack.go
```diff
@@ -13,6 +13,8 @@ import (
 	"sync"
 	"time"
 
+	"strconv"
+
 	"golang.org/x/net/http2"
```
tsenart: Please run goimports to automatically group packages in blocks

lib/attack.go
```diff
@@ -348,6 +350,8 @@ func (a *Attacker) hit(tr Targeter, name string) *Result {
 		}
 	}()
 
+	tgt.Header.Set("X-Vegeta-Seq", strconv.FormatUint(res.Seq, 10))
+
```
tsenart: I don’t think users should be able to over-ride this header unintentionally, so we should set it after the target is read, which also ensures there’s an http.Header (so we don’t need the allocation).

tsenart:
Ping @jackdbernier

jackdbernier:
I’m planning on coming back to this. I’m currently  on vacation so it might
take a few weeks.

jackdbernier:
Awesome! Thank you so much.

I will take a look at it when I get back to work.

PR #474, opened 2020-02-07
Capture HTTP headers in results

### LINKED ISSUE #303: Capture HTTP response header in vegeta.Result

rs:
Headers are written in wire format. CSV output uses base64 to encode
headers.

Fixes #303

### DIFF
--- a/lib/attack.go
+++ b/lib/attack.go
@@ -1,6 +1,7 @@
 package vegeta
 
 import (
+	"bytes"
 	"context"
 	"crypto/tls"
 	"fmt"
@@ -391,6 +392,10 @@ func (a *Attacker) hit(tr Targeter, name string) *Result {
 	if res.Code = uint16(r.StatusCode); res.Code < 200 || res.Code >= 400 {
 		res.Error = r.Status
 	}
+
+	var hdr bytes.Buffer
+	r.Header.Write(&hdr)
+	res.Headers = hdr.String()
 
 	return &res
 }
--- a/lib/attack.go
+++ b/lib/attack.go
@@ -1,7 +1,6 @@
 package vegeta
 
 import (
-	"bytes"
 	"context"
 	"crypto/tls"
 	"fmt"
@@ -393,9 +392,7 @@ func (a *Attacker) hit(tr Targeter, name string) *Result {
 		res.Error = r.Status
 	}
 
-	var hdr bytes.Buffer
-	r.Header.Write(&hdr)
-	res.Headers = hdr.String()
+	res.Headers = r.Header
 
 	return &res
 }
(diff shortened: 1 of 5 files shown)


lib/attack.go
```diff
@@ -392,5 +393,9 @@ func (a *Attacker) hit(tr Targeter, name string) *Result {
 		res.Error = r.Status
 	}
 
+	var hdr bytes.Buffer
+	r.Header.Write(&hdr)
```
golangcibot: Error return value of `r.Header.Write` is not checked (from `errcheck`)

lib/results.go
```diff
@@ -32,6 +32,7 @@ type Result struct {
 	Body      []byte        `json:"body"`
 	Method    string        `json:"method"`
 	URL       string        `json:"url"`
+	Headers   string        `json:"headers"`
```
tsenart: I think it would be useful to marshal out headers in a structured format in the JSON encoding. Can we make this an http.Header with the intent?

rs: I thought about it but decide against it because:
* The marshaling of headers would be `map[string][]string`, which is not very easy to use
* You lose the ordering of headers that might be interesting for some type of debugging

tsenart: Could we then preserve the ordering with a different type? I'd like to be able to easily process results and their headers in the terminal with `jq` or similar.

How about:

```go
type Header struct { Key, Value string }

type Result struct {
    // ...
    Headers []Header
```

rs: How would you encode that in CSV format?

tsenart: I am not sure. We could just leave as you originally had it, unless you have a better idea :-)

rs: Not really. But then you want a JSON version for JSON and b64 wire format for CSV?

tsenart: Yeah, I think it’s totally fine for the different encodings to represent different fields differently.

lib/results.go
```diff
@@ -217,6 +254,15 @@ func NewCSVDecoder(r io.Reader) Decoder {
 
 		r.Method = rec[9]
 		r.URL = rec[10]
+		if len(rec) > 11 {
```
tsenart: I think if we set `FieldsPerRecord = 12`, then this is always going to be true, or else an error would be returned by read.

> FieldsPerRecord is the number of expected fields per record. If FieldsPerRecord is positive, Read requires each record to have the given number of fields.

tsenart: Addressing this in a follow up commit. Thanks for the contribution!

PR #475, opened 2020-02-08
Add the max-connections option

rs:
The connections option is misleading, as it let you think that no more
than the given number of connections will be established during the
attack, whereas it does only control the maximum amount of idle connection
in the pool, but keep the maximum number of connection unlimited.

The max-connection sets the maximum number of connections per target
host, which is what most people certainly expect the connections option
does.

#### Checklist

### DIFF
--- a/lib/attack.go
+++ b/lib/attack.go
@@ -41,6 +41,9 @@ const (
 	// DefaultConnections is the default amount of max open idle connections per
 	// target host.
 	DefaultConnections = 10000
+	// DefaultMaxConnections is the default amount of connections per target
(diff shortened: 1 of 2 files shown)


lib/attack.go
```diff
@@ -41,6 +41,9 @@ const (
 	// DefaultConnections is the default amount of max open idle connections per
 	// target host.
 	DefaultConnections = 10000
+	// DefaultMaxConnections is the default amount of connections per target
+	// host.
```
```suggestion
	// host. Zero means no limit.
```

PR #481, opened 2020-02-20
Add context to targets and results for implements complex loadtest scenarios

KoteiIto:
#### Background
Hello.
We use this library to implement complex load testing scenarios, such as running multiple APIs sequentially.
At that time, Targets change dynamically from the result of the previous API, but it is difficult to realize with the current function.
Therefore, we added the function to save context.Context in Targets and save it in Results.

implementation looks like this:
```
tr := func(t *vegeta.Target) error {
	ctx := <-ctxListener
	// get value from Context and Serialize Body or build params to create target
	*t = target
	return nil
}

listener := attacker.Attack(tr, pacer, duration, name)
for result := range listener {
	metrics.Add(result)
	encoder.Encoder(result)

	ctx := result.Context()
	// Deserialize result.Body and set value to context.Context
	ctxEmitter <- ctx
}
```
Please consider it.

#### Checklist


blakewatters:
👍 I am in a similar situation and trying to associate context with requests and responses

tsenart:
I don't think you need this code change to support use cases like this one since you can share state between a the go-routine reading results from the attack and your custom `Targeter`.

tsenart:
> 👍 I am in a similar situation and trying to associate context with requests and responses

Would it help if a `Request` would contain its originating `Target`?

PR #495, opened 2020-03-01
pacer: Extend interface with Rate method

### LINKED ISSUE #476: Make rate calculation of the instantaneous rate of attack exportable
#### Proposal

I wish I could get the instantaneous rate of attack. It means make exportable rate function in pacer.go at line 272

#### Background

It would be useful to me to observe the growth of the rate when using linear pacer.

#### Workarounds

I can't find any other workaround

tsenart:
This commit extends the Pacer interface with a Rate method that returns
the instantaneous hit rate (per second) at a given point in time.

Fixes #476

### DIFF
--- a/lib/pacer.go
+++ b/lib/pacer.go
@@ -6,12 +6,17 @@ import (
 	"time"
 )
 
-// A Pacer defines the rate of hits during an Attack by
-// returning the duration an Attacker should wait until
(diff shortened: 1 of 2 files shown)


tsenart:
PTAL @francescomormile and @blakewatters

PR #512, opened 2020-04-22
fix: fix module import path for example

C-ZW:
#### Background

When I try to run the example code, got a import error.
I can't find the v12 in lib forder in repository, after remove /v12 in path, it works.

#### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -762,7 +762,7 @@ import (
   "fmt"
   "time"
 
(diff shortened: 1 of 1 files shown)


tsenart:
`/v12` needs to be there for your code to be compiled with go modules enabled. Since go modules is the official blessed way forward, I don't want to make the old $GOPATH approach front and center.

Issue #477, opened 2020-02-14
Prometheus support


axot:
Hello, it is great to see metrics in prometheus, is it possible to do this?
jaggr is used to plot realtime graph, so I think it should possible to convert it to prometheus format.

Any thoughts?

tsenart:
It would be possible to create a sub command that consumes vegeta attack results (i.e. like vegeta report or vegeta encode) that is a prometheus exporter.

flaviostutz:
## Proposal

What do you think about creating a new command category called "monitor" aside report, encode, dump?

### Monitor

So for monitoring with Prometheus one could do something like

```
GET http://www.google.com" | vegeta attack -duration=5m | vegeta monitor -type='prometheus[8880,0.0.0.0]'
```

* "-type" may be omitted. defaults to 'prometheus'
* "8880" is the Prometheus exporter http port (may be omitted and defaults to 8880)
* "0.0.0.0" is the bind host for the listener (may be omitted and defaults to 0.0.0.0)

After launching the tests, just point Prometheus scraper to http://host:8880/metrics or do something like

```
curl http://host:8880/metrics
```

### Prometheus Metrics

* ```request_seconds_count{method, path, status}```
* ```request_seconds_sum{method, path, status}```

What do you think?

tsenart:
That’s a good idea.

Sent via Superhuman iOS ( https://sprh.mn/?vip=tsenart@gmail.com )

tsenart:
Alternatively we’d instrument attack directly.

Sent via Superhuman iOS ( https://sprh.mn/?vip=tsenart@gmail.com )

flaviostutz:
Much better!

Instrumenting “attack” is great because it avoids parsing stdin contents and performance will be better for large scale distributed deployments.

tsenart:
@flaviostutz: Contributions welcome! :)

flaviostutz:
Sure! :)

varyumin:
Do you have news about prom metrics? I am really looking forward!

flaviostutz:
I implemented it and a PR is being analyzed. Soon we will have it! ;)

shydefoo:
Is there any update on the status of this PR? This feature would be really useful, thanks!

real-zony:
Is there any update on the status of this PR? This is a very useful feature in our DevOps process.

ocodista:
Hi!!

Can you please add a command example on the docs? I've been trying to make it work but simply cant.

PR #534, opened 2020-07-24
Implementing Prometheus Exporter and documentation

flaviostutz:
#### Background

This is a complete initial implementation of Prometheus Exporter support as discussed in #477 

#### Checklist

- [X] Changed or added code is covered by appropriate tests.


flaviostutz:
This is a continuation of PR https://github.com/tsenart/vegeta/pull/526 but decided to create a new PR to cleanup things for a new review.

flaviostutz:
@tsenart Could you please give a check on this PR?

tsenart:
@flaviostutz: I've been swamped with work but saved some time to look at this on Saturday! Sorry.

flaviostutz:
No worries... I know how it is... here with me it's the same... thanks!

Dockerfile
```diff
@@ -0,0 +1,28 @@
...
+ADD go.mod /vegeta
+ADD go.sum /vegeta
+RUN go mod download
+
+#now build source code
+ADD / /vegeta
+
+RUN go test -v ./...
+RUN make vegeta
+# RUN go build -v -o /bin/vegeta
```
tsenart, line 17:
```suggestion
```

Dockerfile
```diff
@@ -0,0 +1,28 @@
...
+RUN apk add make build-base
+
+WORKDIR /vegeta
+
+# cache dependencies
+ADD go.mod /vegeta
+ADD go.sum /vegeta
+RUN go mod download
+
+#now build source code
```
tsenart, line 12:
```suggestion
# now build source code
```

Dockerfile
```diff
@@ -0,0 +1,28 @@
...
+
+RUN go test -v ./...
+RUN make vegeta
+# RUN go build -v -o /bin/vegeta
+
+FROM alpine:3.12.0
+
+ENV TARGET_URL ''
+ENV DURATION '5'
+ENV REQUESTS_PER_SECOND '5'
```
tsenart, line 23: Rather than using environment variables, which would require us to map each flag to an env var, can we just expose the vegeta binary as a command so that we're able to pass flags directly to `docker run` ?

README.md
```diff
@@ -94,6 +94,14 @@ attack command:
     	Attack name
   -output string
     	Output file (default "stdout")
+  -prom-bind
```
tsenart, line 97: What do you think of replacing these four flags with a single `-prometheus-url`? Example `-prometheus-url=http://0.0.0.0:8880/metrics`. It being defined would imply it's enabled. By default, it'd be empty.

README.md
```diff
@@ -727,6 +735,8 @@ It'll read and sort them by timestamp before generating reports.
 vegeta report *.bin
 ```
 
+Another way to gather results in distributed tests is to use the built-in Prometheus Exporter and configure a Prometheus Server to get test results from all Vegeta instances. See `attack` option "prom-enable" for more details and a complete example at section "Prometheus Exporter Support"
```
tsenart, line 738:
```suggestion
Another way to gather results in distributed tests is to use the built-in Prometheus Exporter and configure a Prometheus Server to get test results from all Vegeta instances. See `attack` option "prom-enable" for more details and a complete example in the section "Prometheus Exporter Support"
```

README.md
```diff
@@ -803,6 +813,65 @@ $ ulimit -u # processes / threads
 
 Just pass a new number as the argument to change it.
 
+## Prometheus Support
+
+Vegeta has a built-in Prometheus Exporter that may be enabled during "attacks" so that you can point any Prometheus instance to Vegeta instances and get some metrics about http requests performance and about Vegeta process itself.
```
tsenart, line 818:
```suggestion
Vegeta has a built-in Prometheus Exporter that may be enabled during "attacks" so that you can point any Prometheus instance to Vegeta instances and get some metrics about http requests performance and about the Vegeta process itself.
```

README.md
```diff
@@ -803,6 +813,65 @@ $ ulimit -u # processes / threads
...
+
+To enable Prometheus Exporter on command line, see flags with "prom-" prefix.
```
tsenart, line 820:
```suggestion
To enable the Prometheus Exporter on the command line, see flags with "prom-" prefix.
```

README.md
```diff
@@ -803,6 +813,65 @@ $ ulimit -u # processes / threads
...
+To enable Prometheus Exporter on lib usage, add Option "PrometheusEnable" and/or "PrometheusSettings".
```
tsenart, line 821:
```suggestion
To enable the Prometheus Exporter in library, add Option "PrometheusEnable" and/or "PrometheusSettings".
```

README.md
```diff
@@ -803,6 +813,65 @@ $ ulimit -u # processes / threads
...
+
+Prometheus HTTP endpoint will be available only during the lifespan of an "attack" and will be closed right after the attack is finished.
```
tsenart, line 823:
```suggestion
A Prometheus HTTP endpoint will be available only during the lifespan of an "attack" and will be closed right after the attack is finished.
```

README.md
```diff
@@ -803,6 +813,65 @@ $ ulimit -u # processes / threads
...
+
+The following metrics are exposed:
+
+* `request_bytes_in` - bytes count received from targeted servers by "url", "method" and "status"
+* `request_bytes_out` - bytes count sent to targeted server by "url", "method" and "status"
+* `request_seconds` - histogram with request latency and counters by "url", "method" and "status"
```
tsenart, line 829: Can we derive actual request rate from these metrics so that we could plot it or do we need to capture it separately?

lib/prom/prom.go
```diff
@@ -0,0 +1,121 @@
...
+	"time"
+
+	"github.com/gorilla/mux"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
+	"github.com/prometheus/client_golang/prometheus/promhttp"
+	vegeta "github.com/tsenart/vegeta/v12/lib"
+)
+
+//PrometheusMetrics vegeta metrics observer with exposition as Prometheus metrics endpoint
```
tsenart, line 16: Please fix the spacing between the // and the start of the comment text everywhere.

lib/prom/prom.go
```diff
@@ -0,0 +1,121 @@
...
+type PrometheusMetrics struct {
```
tsenart, line 17: Since the package name is already `prom`, I think we can get rid of the `Prometheus` prefix everywhere, so that it reads `prom.Metrics`

lib/prom/prom.go
```diff
@@ -0,0 +1,121 @@
...
+	prometheus.Unregister(pm.requestSecondsHistogram)
+	prometheus.Unregister(pm.requestBytesInCounter)
+	prometheus.Unregister(pm.requestBytesOutCounter)
+	prometheus.Unregister(pm.requestFailCounter)
+	return pm.listenPort.Close()
+}
+
+//Observe register metrics about hit results
+func (pm *PrometheusMetrics) Observe(res *vegeta.Result) {
+	pm.requestBytesInCounter.WithLabelValues(res.Method, res.URL, fmt.Sprintf("%d", res.Code)).Add(float64(res.BytesIn))
```
tsenart, line 115: Replace usage of `fmt.Sprintf` with `strconv.FormatUint` (faster).

lib/prom/prom.go
```diff
@@ -0,0 +1,121 @@
...
+		Name: "request_fail_count",
+		Help: "Internal failures that prevented a hit to the target server",
+	}, []string{
+		"method",
+		"url",
+		"message",
+	})
+
+	//setup prometheus metrics http server
+	router := mux.NewRouter()
```
tsenart, line 85: I don't think there's a need to add mux as a dependency? We can use the native on in the `http` package.

lib/prom/prom.go
```diff
@@ -0,0 +1,121 @@
...
+	router.Handle(metricsPath, promhttp.Handler())
+
+	listen := fmt.Sprintf("%s:%d", bindHost, bindPort)
+	lp, err := net.Listen("tcp", listen)
```
tsenart, line 89: Any particular reason to create the socket independently? Why not use the simpler:

```
s := &http.Server{
	Addr:           ":8080",
	Handler:        myHandler,
}
s.ListenAndServe()
```

lib/prom/prom.go
```diff
@@ -0,0 +1,121 @@
...
+//   - bindPort: port to bind the listening socket to
+//   - metricsPath: http path that will be used to get metrics
+// For example, after using NewPrometheusMetricsWithParams("0.0.0.0", 8880, "/metrics"),
+// during an "attack" you can call "curl http://127.0.0.0:8880/metrics" to see current metrics.
+// This endpoint can be configured in scrapper section of your Prometheus server.
+func NewPrometheusMetricsWithParams(bindHost string, bindPort int, metricsPath string) (PrometheusMetrics, error) {
+
+	pm := PrometheusMetrics{}
+
+	pm.requestSecondsHistogram = promauto.NewHistogramVec(prometheus.HistogramOpts{
```
tsenart, line 47: Using promauto means we write to the global package level Prometheus registry. This may have unexpected effects when vegeta is used as a library. Is there a way to be explicit about adding the metrics to a specific registry?

lib/results_test.go
```diff
@@ -102,6 +103,9 @@ func TestResultEncoding(t *testing.T) {
 					URL: rapid.String().Draw(t, "url").(string),
 				}
 
+				want.Error = strings.ReplaceAll(want.Error, "\n", "")
```
tsenart, line 106: Can you explain this change?

flaviostutz: rapid uses random seeds at each test and during fuzzy tests sometimes it uses "\n" chars in strings but decoder misplaces it by "\r\n". Break lines in URL strings is unlikely to occur and rapid doesn't have a way to fix initial seed value to avoid those random errors (https://github.com/flyingmutant/rapid/issues/14).

lib/prom/prom.go
flaviostutz: Sure.

lib/prom/prom.go
flaviostutz: We create the socket independently so that we can close it when context.Context finishes. There is no way to cancel the server with the simpler version.

tsenart: I think you can use https://golang.org/pkg/net/http/#Server.Shutdown

Dockerfile
flaviostutz: done

Dockerfile
flaviostutz: done

Dockerfile
flaviostutz: done

README.md
flaviostutz: done

README.md
flaviostutz: done

lib/prom/prom.go
flaviostutz: done

lib/prom/prom.go
flaviostutz: done

lib/prom/prom.go
flaviostutz: done

flaviostutz:
@tsenart could you please take a look at this PR when you have some time? I think I managed to do all the requested changes.

README.md
```diff
@@ -803,6 +811,62 @@ $ ulimit -u # processes / threads
 
 Just pass a new number as the argument to change it.
 
+## Prometheus Support
+
+Vegeta has a built-in Prometheus Exporter that may be enabled during "attacks" so that you can point any Prometheus instance to Vegeta instances and get some metrics about http requests performance and about the Vegeta process itself.
+
+To enable the Prometheus Exporter on the command line, see flags with "prom-" prefix.
```
tsenart, line 818:
```suggestion
To enable the Prometheus Exporter on the command line, use the "prometheus-url" flag.
```

README.md
tsenart: Thoughts?

lib/prom/prom_test.go
```diff
@@ -0,0 +1,85 @@
...
+	r.Code = 500
+	r.Error = "REQUEST FAILED"
+	pm.Observe(r)
+
+	resp, err = http.Get("http://localhost:8880")
+	assert.Nil(t, err, "Error calling prometheus metrics. err=%s", err)
+	assert.Equal(t, 200, resp.StatusCode, "Status code should be 200")
+
+	data, err = ioutil.ReadAll(resp.Body)
+	assert.Nil(t, err, "Error calling prometheus metrics. err=%s", err)
```
tsenart, line 78: Please don't introduce the testify dependency.

lib/prom/prom.go
```diff
@@ -0,0 +1,134 @@
...
+
+// NewPrometheusMetricsWithParams creates a new Prometheus Metrics to Observe attack results and expose metrics
+// For example, after using NewPrometheusMetricsWithParams("0.0.0.0:8880"),
+// during an "attack" you can call "curl http://127.0.0.0:8880" to see current metrics.
+// This endpoint can be configured in scrapper section of your Prometheus server.
+func NewPrometheusMetricsWithParams(bindURL string) (*PrometheusMetrics, error) {
+
+	//parse bind url elements
+	re := regexp.MustCompile("(.+):([0-9]+)")
+	rr := re.FindAllStringSubmatch(bindURL, 3)
```
tsenart, line 40: There are built-in functions for this. Use `url.Parse` and `net.SplitHostPort`

lib/prom/prom.go
tsenart: Please do this.

tsenart (approved):
Looks good! Just a few leftover changes / inconsistencies. Thank you 🙇

lib/prom/prom.go
flaviostutz: ok. thanks for the advice!

Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+
+# cache dependencies
+ADD go.mod /vegeta
+ADD go.sum /vegeta
+RUN go mod download
+
+# now build source code
+ADD / /vegeta
+
+RUN go test -v ./...
```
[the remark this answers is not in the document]
flaviostutz: ok. removed

README.md
flaviostutz: fixed

lib/prom/prom_test.go
flaviostutz: removed

README.md
tsenart: @flaviostutz: Pending input here.

go.sum
```diff
@@ -28,34 +57,110 @@ github.com/gonum/stat v0.0.0-20181125101827-41a0da705a5b h1:fbskpz/cPqWH8VqkQ7LJ
...
+github.com/prometheus/procfs v0.1.3 h1:F0+tqvhOksq22sc6iCHF5WGlWjdwj92p0udFh1VFBS8=
+github.com/prometheus/procfs v0.1.3/go.mod h1:lV6e/gmhEcM9IjHGsFOCxxuZ+z1YqCvr4OA4YeYWdaU=
+github.com/sirupsen/logrus v1.2.0/go.mod h1:LxeOpSwHxABJmUn/MG1IvRgCAasNZTLOkJPxbbu5VWo=
+github.com/sirupsen/logrus v1.4.2/go.mod h1:tLMulIdttU9McNUspp0xgXVQah82FyeX6MwdIuYE2rE=
 github.com/streadway/quantile v0.0.0-20150917103942-b0c588724d25 h1:7z3LSn867ex6VSaahyKadf4WtSsJIgne6A1WLOAGM8A=
 github.com/streadway/quantile v0.0.0-20150917103942-b0c588724d25/go.mod h1:lbP8tGiBjZ5YWIc2fzuRpTaz0b/53vT6PEs3QuAWzuU=
+github.com/stretchr/objx v0.1.0/go.mod h1:HFkY916IF+rwdDfMAkV7OtwuqBVzrE8GR6GFx+wExME=
+github.com/stretchr/objx v0.1.1/go.mod h1:HFkY916IF+rwdDfMAkV7OtwuqBVzrE8GR6GFx+wExME=
+github.com/stretchr/testify v1.2.2/go.mod h1:a8OnRcib4nhh0OaRAV+Yts87kKdq0PP7pXfy6kDkUVs=
+github.com/stretchr/testify v1.3.0/go.mod h1:M5WIy9Dh21IEIfnGCwXGc5bZfKNJtfHm1UVUgZn+9EI=
```
tsenart, line 113: @flaviostutz: Why is this still here?

lib/prom/prom.go
tsenart: @flaviostutz: Pending.

lib/prom/prom.go
tsenart: @flaviostutz: Pending.

README.md
hfuss: With prometheus you typically keep metrics very raw as counters and derive rates from them using PromQL.

This way if you had multiple Vegeta's running at once you could calculate the rates in aggregate or individually across multiple pods.

README.md
```diff
@@ -803,6 +807,62 @@ $ ulimit -u # processes / threads
 
 Just pass a new number as the argument to change it.
 
+## Prometheus Support
+
+Vegeta has a built-in Prometheus Exporter that may be enabled during "attacks" so that you can point any Prometheus instance to Vegeta instances and get some metrics about http requests performance and about the Vegeta process itself.
```
hfuss, line 812: While I like this for a lot reasons, I think vegeta should be exporting metrics to a Prometheus push gateway rather than being scraped directly by prometheus.

Since vegeta isn't a long lived process you'll have race conditions where Prometheus might not scrape the last bit of attack data before vegeta shuts down.

Prometheus [best practices docs](https://prometheus.io/docs/practices/pushing/) say you should use push gateway for "service-level batch jobs" which is what I think vegeta would qualify as:
> Usually, the only valid use case for the Pushgateway is for capturing the outcome of a service-level batch job. A "service-level" batch job is one which is not semantically related to a specific machine or job instance (for example, a batch job that deletes a number of users for an entire service). Such a job's metrics should not include a machine or instance label to decouple the lifecycle of specific machines or instances from the pushed metrics. This decreases the burden for managing stale metrics in the Pushgateway. See also the best practices for monitoring batch jobs.

hfuss: I bring this up, bc I think it should instead have, or at least additionally offer, a report type of `prom` which will then export metrics from a report to a prometheus push gateway.

I am currently writing that for `datadog` rather than Prometheus but the idea is the same. The advantage of the report is that one could take an existing Vegeta test result and push into a metrics store rather than having to run a new test.

And the user could publish the results to a Prometheus and a DataDog and wherever else they needed with the reporting decoupled from the attacking.

lucmichalski:
Hi guys, 

Hope you are all well !

Just wanted to know the current status of this PR as it sounds awesome for load testing any webapp.

What is missing to merge it ?

Cheers,
Luc

leosunmo:
As someone who uses Vegeta as a library for testing/benchmarking/loadtesting, this is fantastic! Having to manually implement OpenCensus middleware for HTTP is a bit of a pain.

Just a note, is there a way to customise the latency histogram buckets? It looked like they were hardcoded to me, and since `requestSecondsHistogram` is not exposed I can't change it.

dntosas:
hola people!

any news on this one? seems this functionality is widely needed and a lot of thanks for your nice job cc @flaviostutz @tsenart

spacez320:
@hfuss @flaviostutz I'm wondering what the state of the PR is and if it's still possible to implement Prometheus support.

I agree that normally an ad-hoc job like a Vegeta process could benefit from just delivering results to Pushgateway. I also think creating an exporter could be useful if you're going to run Vegeta for a long time and want a simpler integration.

I'm preparing to use Vegeta for some long-running (hours-days) load-testing jobs and wondered if I could help at all with Prometheus integration one way or the other.

- Do we think it's still possible to create the exporter?
- Should we instead (or also) make it easy to transmit to Pushgateway via `vegeta report` or something?

daluu:
Depending on use case, I would advise against the PushGateway approach except as a last resort. In my efforts to use it, observed that metrics over the push gateway have no TTL and persist forever, and the metric's value only updates if you send new values over time with the exact same labels/dimensions. If the labels change over time (different instance, host, etc.), you'll just get new metrics instead and both old and new metrics persist in the system over time.

So when say viewing the metrics in grafana, you have a never ending set of timeseries, rather than timeseries with filled lines for whenever metrics are sent to gateway, and missing data on chart for intervals where no data is sent to push gateway. I wanted and expected the latter but got the former. For ad-hoc and periodic load testing, I'd assume people would want the latter. The former is more suitable if say you ran vegeta for monitoring of a test instance that never changes, same service you are monitoring over time.

For what I want to do, if you go with push gateway, you may need something like this forked version instead that offers TTL to the push gateway metrics: https://github.com/dinumathai/pushgateway

So I think the exporter route is good. Another option to consider for alternative to exporter, is instead of an interface for prometheus to scrap the metrics from vegeta, can look into the Prometheus remote write protocol/feature, as a way to push metrics to prometheus. We're looking into and testing the remote write out at our organization, I'm not aware of the specific implementation or how well it's performing at the moment. But I wouldn't know for case of vegeta, which would be more optimal from a performance standpoint since vegeta needs to both generate the load (test) and also expose or push the metrics.

tsenart:
@daluu: Makes sense. I think the Prometheus exporter should then be implemented as a vegeta sub command, like `vegeta prom-export`, which you'd use with `vegeta attack ... | tee results.bin | vegeta prom-export`

README.md
flaviostutz: Yes you are right. Rate is done by the Prometheus server while it queries the timeseries database after data ingestion.

(Nice that this project is being reactivated!)

lib/prom/prom.go
flaviostutz: I will work on these on the weekend

tsenart: Whoop whoop! Did you see https://github.com/tsenart/vegeta/pull/534#issuecomment-1630248104 ?

flaviostutz:
As discussed in #477 we decided to instrument attack because then we don't need to parse stdin and it would be simpler/more performant. 

If we decide to change this at this point of the PR actually we should cancel it and start a new branch because a lot of things will be different.

I would recommend us to finish this PR with the initial requirements because it's almost there and if necessary in the future we discuss better about creating a reporter/push gateway/direct write version of it.

What do you think?

tsenart:
Ah, I had forgotten that discussion. It was a long time ago!

Thinking about it now, having the attack command expose a web server handler that a Prometheus instance can scrape is good for performance and reducing moving pieces in distributed attacks.

But for interactive use and debugging past attacks I think we should still have a sub command that exporta saved results as Prometheus metrics.

So, the answer is I think we need both, and they should share as much code as possible. The only difference between doing it in attack and doing it in another sub-command is that in attack we can observe the metric without decoding the result first.

I suggest we introduce a lib/prom package where all of this is encapsulated and modularized.

Also, very much up for hacking on this together. So let me know how you'd want to go about it.

lib/prom/prom.go
flaviostutz: done

lib/prom/prom.go
flaviostutz: done

lib/prom/prom.go
flaviostutz: done

go.sum
flaviostutz: I removed all dependencies to testify in prom lib, but in go.sum there is an existing testify dependency in master already and I don't know where it comes from. go.mod doesn't declare it. probably an indirect dependency somewhere (but it already exists before my PR)

flaviostutz:
> Ah, I had forgotten that discussion. It was a long time ago!
> 
> Thinking about it now, having the attack command expose a web server handler that a Prometheus instance can scrape is good for performance and reducing moving pieces in distributed attacks.
> 
> But for interactive use and debugging past attacks I think we should still have a sub command that exporta saved results as Prometheus metrics.
> 
> So, the answer is I think we need both, and they should share as much code as possible. The only difference between doing it in attack and doing it in another sub-command is that in attack we can observe the metric without decoding the result first.
> 
> I suggest we introduce a lib/prom package where all of this is encapsulated and modularized.
> 
> Also, very much up for hacking on this together. So let me know how you'd want to go about it.

This PR is already creating a lib/prom package. I just reviewed and fixed all comments that were open (after 3y lol!). Please do another review round and mark comments as resolved if they are ok.

flaviostutz:
I squashed the messy commits (various upstream merges and small changes) from my branch into one and applied the PGP signatures so you can merge it to master 😁

lib/prom/prom_test.go
```diff
@@ -0,0 +1,132 @@
...
+
+	resp, err = http.Get("http://localhost:8880")
+	if err != nil {
+		t.Errorf("Error calling prometheus metrics. err=%s", err)
+	}
+	if resp.StatusCode != 200 {
+		t.Errorf("Status code should be 200")
+	}
+
+	data, err = ioutil.ReadAll(resp.Body)
```
tsenart, line 118: Replace `ioutil` with `io` everywhere please.

tsenart (commented):
Thank you for moving this forward! Out of all my comments, let me know what you want to work on. I'm happy to address my own feedback on top of your changes. I'll also implement the `prom` sub-command that I mentioned.

README.md
tsenart: Note to self: When `-prometheus-addr` is set and we start an HTTP server for prometheus to scrape, make sure that all metrics are scrapped before exiting from the program, even after the attack finished per se.

lib/results_test.go
flaviostutz: removed

docker-compose.yml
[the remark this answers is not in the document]
flaviostutz: done

grafana.json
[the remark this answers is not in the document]
flaviostutz: done

lib/prom/prom.go
```diff
@@ -0,0 +1,127 @@
...
+}
+
+// Observe register metrics about hit results
+func (pm *Metrics) Observe(res *vegeta.Result) {
+	code := strconv.FormatUint(uint64(res.Code), 10)
+	pm.requestBytesInCounter.WithLabelValues(res.Method, res.URL, code).Add(float64(res.BytesIn))
+	pm.requestBytesOutCounter.WithLabelValues(res.Method, res.URL, code).Add(float64(res.BytesOut))
+	pm.requestSecondsHistogram.WithLabelValues(res.Method, res.URL, code).Observe(float64(res.Latency) / float64(time.Second))
+	if res.Error != "" {
+		pm.requestFailCounter.WithLabelValues(res.Method, res.URL, res.Error)
```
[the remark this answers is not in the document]
flaviostutz: done

lib/prom/prom.go
```diff
@@ -0,0 +1,127 @@
...
+	}
+
+	pm := &Metrics{
+		registry: prometheus.NewRegistry(),
+	}
+
+	pm.requestSecondsHistogram = prometheus.NewHistogramVec(prometheus.HistogramOpts{
+		Name:    "request_seconds",
+		Help:    "Request latency",
+		Buckets: []float64{0.1, 0.2, 0.5, 1.0, 2.0, 5.0, 10.0, 20, 50},
```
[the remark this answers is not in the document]
flaviostutz: done

lib/prom/prom.go
```diff
@@ -0,0 +1,127 @@
...
+// NewMetrics same as NewMetricsWithParams with default params:
+func NewMetrics() (*Metrics, error) {
+	return NewMetricsWithParams("http://0.0.0.0:8880")
+}
+
+// NewMetricsWithParams creates a new Prometheus Metrics to Observe attack results and expose metrics
+// For example, after using NewMetricsWithParams("http://0.0.0.0:8880"),
+// during an "attack" you can call "curl http://127.0.0.0:8880" to see current metrics.
+// This endpoint can be configured in scrapper section of your Prometheus server.
+func NewMetricsWithParams(bindURL string) (*Metrics, error) {
```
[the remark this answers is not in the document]
flaviostutz: changed

lib/prom/prom.go
```diff
@@ -0,0 +1,127 @@
...
+		pm.srv.ListenAndServe()
+	}()
+
+	return pm, nil
+}
+
+// Close shutdown http server exposing Prometheus metrics and unregister
+// all prometheus collectors
+func (pm *Metrics) Close() error {
+	prometheus.Unregister(pm.requestSecondsHistogram)
```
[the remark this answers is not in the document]
flaviostutz: good catch! done

lib/prom/prom.go
```diff
@@ -0,0 +1,127 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("Invalid bindURL %s. Must be in format 'http://0.0.0.0:8880'. err=%s", bindURL, err)
+	}
+	bindHost, bindPort, err := net.SplitHostPort(p.Host)
+	if err != nil {
+		return nil, fmt.Errorf("Invalid bindURL %s. Must be in format 'http://0.0.0.0:8880'. err=%s", bindURL, err)
+	}
+
+	pm := &Metrics{
+		registry: prometheus.NewRegistry(),
```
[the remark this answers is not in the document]
flaviostutz: I was trying to avoid mixing prometheus with attack.go at first, making it work only on cli so we don't change "attack" before we check the stability of prometheus metrics.

lib/prom/prom_test.go
flaviostutz: oh, this is new for me. done

docker-compose.yml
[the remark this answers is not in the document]
flaviostutz: couldn't be done because docker-compose.yml have to have Dockerfile in a child dir and Dockerfile has to be in root because it compiles the entire vegeta lib in the container

prometheus-sample.png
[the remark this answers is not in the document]
flaviostutz: Yes. I will document prometheus query examples for that too

lib/prom/prom.go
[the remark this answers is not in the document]
tsenart: Since `lib/prom` is a public package, we need to support library usage as well. We don't need to change `lib/attack.go`, only `attack.go`, where you'd instantiate the server with Prometheus handler. So this type `Metrics` would take in the `prometheus.Registry` and register it's metrics in it, but would not serve anything.

lib/prom/prom.go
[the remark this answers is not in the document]
tsenart: Forgot to push?

docker-compose.yml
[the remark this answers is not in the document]
tsenart: Then I'd move the Dockerfile to lib/prom as well.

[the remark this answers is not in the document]
flaviostutz: Yes. I am preparing to push all changes at all.

lib/prom/prom.go
```diff
@@ -0,0 +1,127 @@
...
+	pm.requestSecondsHistogram = prometheus.NewHistogramVec(prometheus.HistogramOpts{
+		Name:    "request_seconds",
+		Help:    "Request latency",
+		Buckets: []float64{0.1, 0.2, 0.5, 1.0, 2.0, 5.0, 10.0, 20, 50},
+	}, []string{
+		"method",
+		"url",
+		"status",
+	})
+	pm.registry.MustRegister(pm.requestSecondsHistogram)
```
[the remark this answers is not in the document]
flaviostutz: As now we can receive existing registries, we can have errors, so I am changing it :)

attack.go
```diff
@@ -56,6 +58,7 @@ func attackCmd() command {
 	fs.Var(&opts.laddr, "laddr", "Local IP address")
 	fs.BoolVar(&opts.keepalive, "keepalive", true, "Use persistent connections")
 	fs.StringVar(&opts.unixSocket, "unix-socket", "", "Connect over a unix socket. This overrides the host address in target URLs")
+	fs.StringVar(&opts.promURL, "prometheus-url", "", "Enable Prometheus metrics with specific bind parameters in format [bind ip]:[bind port]. Example: 0.0.0.0:8880")
```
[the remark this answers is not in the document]
flaviostutz: done

lib/prom/prom.go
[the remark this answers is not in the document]
flaviostutz: I separated HTTP from metrics and started HTTP server from attack.go, but created an utility function in prom lib to help with this, so that we can use the same piece of code when creating the prom command in the future.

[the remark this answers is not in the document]
flaviostutz: Now it's more clear that someone can use the lib to prepare their own Prom Registry with Vegeta Metrics.

docker-compose.yml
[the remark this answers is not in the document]
flaviostutz: Cannot move Dockerfile to lib/prom because I compile vegeta, so it needs access to the entire repo from root. Keeping it on root

flaviostutz:
@tsenart Pushed everything I changed here. In summary I tried to resolve all comments, so please do another round of checks.

As this PR is too big already, I would advise us to merge it as soon as possible and build the "prom" command in another one.

docker-compose.yml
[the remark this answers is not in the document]
tsenart: I'm merging as is but then following up myself on this. I'd rather have a Docker container built in CI as part of the release process that uses the built binaries in there.

tsenart:
Thank you for your great effort on this <3 Landed in 81403a61f75e655a8bfe2991a58a58defe5a2142. Opened https://github.com/tsenart/vegeta/issues/637 which I'll work on next!

fasibio:
@daluu https://github.com/tsenart/vegeta/pull/534#issuecomment-1629943731
to this point: 
` have no TTL and persist forever, and the metric's value only updates if you send new values over time with the exact same labels/dimensions`

As Info I use a PushGateway and delete at the end all metrics based by vegta so i see no problem with pushgateway: 
```golang
type PushGatewayRegister struct {
	Pusher *push.Pusher
}

// MustRegister implements prometheus.Registerer.
func (p *PushGatewayRegister) MustRegister(c ...prometheus.Collector) {
	for _, v := range c {
		p.Pusher = p.Pusher.Collector(v)
	}
}

// Register implements prometheus.Registerer.
func (p *PushGatewayRegister) Register(c prometheus.Collector) error {
	p.Pusher = p.Pusher.Collector(c)
	return nil
}

// Unregister implements prometheus.Registerer.
func (p *PushGatewayRegister) Unregister(c prometheus.Collector) bool {
	return true
}


func LoadTest() {
      metrics := prom.NewMetrics()

	pusher := push.New("url", "job")
	_ := metrics.Register(&PushGatewayRegister{Pusher: pusher})
       for res := range attacker.Attack(targeter.Targeter, rate, duration, "Big Bang!") {
		metrics.Observe(res)
		err = pusher.Push()
		if err != nil {
			log.Println(fmt.Errorf("push error %w", err))
		}
	}
	_ = pusher.Delete() <- thats the point... all pusher created metrics will be removed from pushgateway
}



```

daluu:
@fasibio 

> As Info I use a PushGateway and delete at the end all metrics based by vegta so i see no problem with pushgateway:

As long as that works out in the end. Then yes should be fine. I would advise to please do test that this works as anticipated, by sending some metrics, then stop sending anymore for some time e.g. over 15 minutes to an hour or more, and verifying that grafana (or equivalent data viewer) properly renders the metric as discrete data points for when metrics were sent rather than an extrapolated line that continues through current time even though no more metrics were sent.

Expecting how something works in theory is different from actually validating it with some testing. So as long as this special handling in push gateway has been confirmed to work as expected with testing, then sounds good.

Note, maybe this approach works better when you have more access to interface to the push gateway, the library/interface we were using at the time, I'm not aware if it had a way to "delete" metrics on the push gateway, we only sent metrics to it.

> but change my mind

I'm not sure what is meant here, it is a little vague for interpretation. Did you mean to say, unless someone can convince you to change your mind otherwise, pushgateway works fine for you? Or did you mean despite what you mentioned, you have decided to change your mind about using push gateway approach?

fasibio:
@daluu 
I use `"github.com/prometheus/client_golang/prometheus/push"` to handle `pusher.Delete()` at the end of the test. 
And so he remove all created Metrics by the same `pusher` object. 

My Grafana graph looks like this: 
image

As you see only at the time of attack there are data. So i think to advise against pushgateway (see readme) is incorrect.

daluu:
> So i think to advise against pushgateway (see readme) is incorrect.

Yes, makes sense, I take back my prior advice, but with the caveat/warning that provided the user has properly utilized the push gateway logic. Because if you omit the delete step at the end, I believe you will run into the concern I previously mentioned, one can try to confirm it. So if we're building a solution and documentation here, need to account for that to ensure proper successful deployment.

@fasibio, curious what made you issue delete at the end of pushing metrics? Somehow you were aware of the need for this (or the issue when you don't delete), or you came across it from trial & error, or found it documented somewhere? Because unless I overlooked the documentation/example code, as far as I can recall I don't recall seeing the documentation or example code indicating to user to issue a delete after pushing out metrics. It's not so intuitive to me how the push gateway was designed, as one would think the push (and pull) model is discrete - you send/poll data, you get data. when no push or poll is occurring, then there should be no data - but the push gateway still holds on to it for continuous forwarding to prometheus if you don't clear it out specifically when the metric values don't change.

fasibio:
Simple I follow the pushgateway "Use it" (no go specific) and there the CURL delete command is part of. 
https://github.com/prometheus/pushgateway?tab=readme-ov-file#use-it . 
And that is all. 

@tsenart see discussion, might make sense to update readme.

PR #562, opened 2021-03-29
Add callback to attacker

bgweber:
#### Background

Adds an optional callback function to the attacker that is called synchronously after each hit.

#### Checklist

### DIFF
--- a/lib/attack.go
+++ b/lib/attack.go
@@ -30,6 +30,7 @@ type Attacker struct {
 	seq        uint64
 	began      time.Time
 	chunked    bool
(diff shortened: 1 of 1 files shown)


tsenart:
You can read from the results channel and act on each result there. Why do we *also* need a callback?

bgweber:
If we can get this from the channel, then this is not needed. Can you share an example of this approach?

tsenart:
https://github.com/tsenart/vegeta

bgweber:
Thanks, closing.

PR #564, opened 2021-04-13
Set body to nil when request is GET

### LINKED ISSUE #556: Target creates a NewReader regardless of underlying Body
#### Version and Runtime

```
v12
```

#### Expected Behaviour
`Target` struct should consider the request method before passing `bytes.NewReader` to `http.NewRequest` as some REST client might reject non-nil body for certain methods (for example `GET`). 

Thus, if I were to use such a client by passing it to vegeta using the `Client` method it would reject my requests.

#### Actual Behaviour
`Target` always calls `NewReader` regardless of the method used in the `Targeter`;
```
func (t *Target) Request() (*http.Request, error) {
	req, err := http.NewRequest(t.Method, t.URL, bytes.NewReader(t.Body))
	if err != nil {
		return nil, err
	}
	for k, vs := range t.Header {
		req.Header[k] = make([]string, len(vs))
		copy(req.Header[k], vs)
	}
	if host := req.Header.Get("Host"); host != "" {
		req.Host = host
	}
	return req, nil
}
```

#### Steps to Reproduce

1. Step 1
1. Step N

#### Additional Context

cc: @jhernand @kurtismulins

chentex:
Update test scenarios

Signed-off-by: Vicente Zepeda Mas <vzepedam@redhat.com>

#### Background

`vegeta` Targeter always sets a body, even if it was set to nil, for any type of requests.

Some APIs have strong restrictions and fail when encountering a body in `GET` request.

Fix #556 

#### Checklist

### DIFF
--- a/lib/targets.go
+++ b/lib/targets.go
@@ -31,9 +31,12 @@ type Target struct {
 // Request creates an *http.Request out of Target and returns it along with an
 // error in case of failure.
 func (t *Target) Request() (*http.Request, error) {
-	req, err := http.NewRequest(t.Method, t.URL, bytes.NewReader(t.Body))
+	req, err := http.NewRequest(t.Method, t.URL, nil)
 	if err != nil {
 		return nil, err
+	}
+	if req.Method != http.MethodGet {
+		req.Body = ioutil.NopCloser(bytes.NewReader(t.Body))
 	}
 	for k, vs := range t.Header {
 		req.Header[k] = make([]string, len(vs))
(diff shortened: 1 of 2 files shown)


lib/targets.go
```diff
@@ -31,10 +31,13 @@ type Target struct {
 // Request creates an *http.Request out of Target and returns it along with an
 // error in case of failure.
 func (t *Target) Request() (*http.Request, error) {
-	req, err := http.NewRequest(t.Method, t.URL, bytes.NewReader(t.Body))
+	req, err := http.NewRequest(t.Method, t.URL, nil)
 	if err != nil {
 		return nil, err
 	}
+	if req.Method != http.MethodGet {
```
tsenart, line 38: This is a backwards incompatible change. I'll push a follow-up commit that sets no body if and only if t.Body is empty or nil, not if the request method is GET.

PR #566, opened 2021-04-28
Add report short output

muff1nman:
#### Background

Error output can get lengthy and is sometimes not useful/desired.

#### Checklist


tsenart:
Changed my mind about this. I think adding flag like this is a pandora's box of complexity, since others will want slightly different output options or tweaks. Instead, I'll add an example on how to filter out the error set with awk.

```
vegeta report | awk '/Error Set:/ {exit} {print}'
```

PR #567, opened 2021-05-03
feat: adding a new param to show url with status code errors

andersonrostirolla:
#### Background

Adding parameter -verbose in report to show all request errors between 400 and 600.

#### Checklist

### DIFF
--- a/lib/metrics.go
+++ b/lib/metrics.go
@@ -38,6 +38,8 @@ type Metrics struct {
 	Success float64 `json:"success"`
 	// StatusCodes is a histogram of the responses' status codes.
 	StatusCodes map[string]int `json:"status_codes"`
+	// RequestsErrors is a requests error (400+ ~ 600-).
+	RequestsErrors []string `json:"requests_errors"`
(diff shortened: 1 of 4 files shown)


lib/metrics.go
```diff
@@ -73,6 +74,11 @@ func (m *Metrics) Add(r *Result) {
 		m.success++
 	}
 
+	if r.Code >= 400 && r.Code < 600 {
+		var codeWithURL string = "Code: " + strconv.Itoa(int(r.Code)) + " - " + r.URL
+		m.RequestsErrors = append(m.RequestsErrors, codeWithURL)
```
tsenart, line 79: This is a potentially huge slice. We could instead group the status code map by URL. I'm gonna experiment with that.

tsenart: OK, I don't think this is solid enough since there are many other request components that we could group by, like the request body, unique set of headers, etc. This is too complex to tack on to this report. You can do your own aggregation with `vegeta encode` and `jq` or `gron`.

Issue #575, opened 2021-07-11
Allow specifying a list of addresses to connect to for a given target host/port


AntonioMA:
#### Proposal
Add a flag to vegeta attack that does something similar to what the --connect-to flag does on curl. The idea would be having something like:

```
echo GET https://mi.internal.server | ./vegeta attack -connect-to my.internal.server:443@10.1.2.3:443,10.1.2.4:443 -insecure
```

that would connect to 10.1.2.3 or 10.1.2.4 when the target host:port is my.internal.server, without hitting the corporate DNS servers. This would also allow running tests locally.

#### Background

Vegeta (go actually) doesn't cache DNS results. That means that any attack with hundreds of thousands of calls end up doing also hundreds of thousands of DNS calls, which is something I don't need or want. On the other hand, our load balancers work distributing calls using SNI, so just connecting by IP would not work.

#### Workarounds

If the host:address where you're connecting doesn't use SNI you could try connecting by IP address. Or you could add the host to your /etc/hosts (or Windows equivalent file) and hope the go net code hits that before hitting the DNS server. Or you could setup a local DNS server to avoid stressing the corporate one.

tsenart:
With DNS caching implemented, I don't think this is still needed.

tsenart:
Implemented by [647dacc](https://github.com/tsenart/vegeta/commit/647dacca770d33913c1b0695b3b71ba261bd341d)

Release: https://github.com/tsenart/vegeta/releases/tag/v12.12.0

PR #576, opened 2021-07-11
[WIP] Add connect-to flag

AntonioMA:
#### Background

This implement the feature requested in https://github.com/tsenart/vegeta/issues/575

#### Checklist

- [X] Changed or added code is covered by appropriate tests.

### DIFF
--- a/attack.go
+++ b/attack.go
@@ -12,12 +12,52 @@ import (
 	"net/http"
 	"os"
 	"os/signal"
+	"regexp"
 	"strings"
 	"time"
 
 	"github.com/tsenart/vegeta/v12/internal/resolver"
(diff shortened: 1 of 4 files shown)


jfreeland:
this would be pretty handy

upgle:
There is quite a difference in performance due to DNS cache. I need this feature too!

tsenart:
Thanks for the contribution. I think I'd rather add the capability of Vegeta caching DNS responses for a configurable duration instead of this. It's my understand that would also solve your problem of overloading your corporate DNS server.

dkegel-fastly:
I still need this; curl doesn't make us set up a special dns server, vegeta shouldn't, either, IMHO.

PR #609, opened 2022-09-20
README: change go get to go install

jinleileiking:
#### Checklist

- [x] Git commit messages conform to [community standards](http://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html).
- [x] Each Git commit represents meaningful milestones or atomic units of work.
- [x] Changed or added code is covered by appropriate tests.

PR #625, opened 2023-06-30
README: fix go install

AlexanderYastrebov:
`go install github.com/tsenart/vegeta@latest` installs 12.7.0

Follows up on #609

#### Checklist


AlexanderYastrebov:
> since there are a few dependencies that need to be installed before

Does it mean that go install does not work anymore? 
It was when created this PR.

tsenart:
It may work but it won't be a proper build with the right metadata, hence the (rather simple) make vegeta target in the Makefile.

PR #632, opened 2023-07-16
DNS caching

tsenart:
This commit introduces DNS caching with the -dns-ttl flag.

Supersedes #576


lib/attack.go
```diff
@@ -263,6 +261,119 @@ func ProxyHeader(h http.Header) func(*Attacker) {
...
+	}
+}
+
+type attack struct {
+	name  string
+	began time.Time
+
+	seqmu sync.Mutex
+	seq   uint64
+}
```
peterbourgon, line 375: If seqmu protects only seq uint64, why not use e.g. [sync/atomic.Uint64](https://pkg.go.dev/sync/atomic#Uint64)?

tsenart: Because it needs to be used to not only protect an increment to seq but also make sure the associated timestamps are in the same order as the sequence number: https://github.com/tsenart/vegeta/blob/9fba76955e42c37f86c7a1c3ebde3b529a145afb/lib/attack.go#L485-L489

peterbourgon: Ah, so given Result values r1 and r2, r1.Seq > r2.Seq implies r1.Timestamp.After(r2.Timestamp)? Very subtle!

tsenart: That's right. Subtle indeed!

PR #634, opened 2023-07-19
Fix: remove ioutil package

fooofei:
#### Background

ioutil package is not recommand to use, instead use io package and os package

#### Checklist

- [ x] Git commit messages conform to [community standards](http://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html).
- [ x] Each Git commit represents meaningful milestones or atomic units of work.
- [ x] Changed or added code is covered by appropriate tests.

### DIFF
--- a/attack.go
+++ b/attack.go
@@ -7,12 +7,12 @@ import (
 	"flag"
 	"fmt"
 	"io"
-	"io/ioutil"
 	"net"
 	"net/http"
 	"os"
 	"os/signal"
 	"strings"
+	"syscall"
 	"time"
 
 	"github.com/tsenart/vegeta/v12/internal/resolver"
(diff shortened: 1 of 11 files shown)


attack.go
```diff
@@ -201,7 +201,7 @@ func attack(opts *attackOpts) (err error) {
 	res := atk.Attack(tr, opts.rate, opts.duration, opts.name)
 	enc := vegeta.NewEncoder(out)
 	sig := make(chan os.Signal, 1)
-	signal.Notify(sig, os.Interrupt)
+	signal.Notify(sig, os.Interrupt, syscall.SIGTERM)
```
tsenart, line 204: This change seems unrelated. Accidental?

tsenart (approved):
LGTM! Thank you. Apart from the signal thing, I'll sign off this commit locally and push to main branch.

attack.go
fooofei: I think SIGTERM is also commonly used, so I add it. Can we add it in this pr? If not, I will remove it.

tsenart: It's fine.

PR #635, opened 2023-07-19
Fix: use atomic.Uint64 instead of  sync.Mutex

fooofei:
#### Background

atomic.Uint64 is better than uint64 +  sync.Mutex, so replace it with atomic.Uint64

#### Checklist

### DIFF
--- a/lib/attack.go
+++ b/lib/attack.go
@@ -13,6 +13,7 @@ import (
 	"net/url"
 	"strconv"
 	"sync"
+	"sync/atomic"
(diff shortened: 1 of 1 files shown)


tsenart (changes requested):
It's my bad for not leaving a good comment around this critical section, but it's necessary for the timestamp calculation to be protected by the same mutex, so that the timestamps are in the same relative order as the sequence numbers. I'd accept a good comment about instead.

Issue #583, opened 2021-12-01
Unusual number of used port & Error "bind: address already in use"


yukha-dw:
#### Question
Hello, when i am trying to test an API, sometimes my vegeta client will spawn a huge swarm of connections eventhough i'm only asking for 2000 connections with a rate of 10k as setup below:
> -connections=2000 -duration=5m -rate=10000 -timeout=1s

I checked using lsof, the highest number it shown is 60k, as high as number of available port? Is it an expected result of busy API or there is a known bug and i just have to restart the vegeta client? Thank you
>$ lsof -n -i | grep "vegeta" -c
>Output > 60439

puneet-traceable:
I was also facing this. Seems like `max-workers` default value is very high. I changed it `no of cores * 4` and it started working well.

tsenart:
If the API is slow to respond then a given port can't be reused until the response comes back. The connections param doesn't limit the number of open connections, only idle connections. Try setting max-workers.

Also make sure to tune your ulimits. Search the reader for ulimit to learn more.

PR #638, opened 2023-07-25
Add optional feature to set the SO_REUSEADDR option before binding th…

fabiorush:
Add an optional parameter to set the SO_REUSEADDR option on the TCP socket before binding
it (default false). Can reduce the amount of "bind: address already in use" errors when doing many
connections to the same server and port combination, like the ones that happened on https://github.com/tsenart/vegeta/issues/583.

### DIFF
--- a/README.md
+++ b/README.md
@@ -136,6 +136,8 @@ attack command:
     	Number of redirects to follow. -1 will not follow but marks as success (default 10)
   -resolvers value
     	List of addresses (ip:port) to use for DNS resolution. Disables use of local system DNS. (comma separated list)
+  -reuseaddr
+    	Set the SO_REUSEADDR socket option (default false)
   -root-certs value
     	TLS root certificate files (comma separated list)
   -session-tickets
@@ -391,6 +393,12 @@ the response is marked as successful.
 
 Specifies custom DNS resolver addresses to use for name resolution instead of
 the ones configured by the operating system. Works only on non Windows systems.
+
+#### `-resolvers`
+
+Specifies whether to set the SO_REUSEADDR option on the TCP socket before binding
+it. Can reduce the amount of "bind: address already in use" errors when doing many
+connections to the same server and port combination.
 
 #### `-root-certs`
 
--- a/attack.go
+++ b/attack.go
@@ -62,6 +62,7 @@ func attackCmd() command {
 	fs.StringVar(&opts.promAddr, "prometheus-addr", "", "Prometheus exporter listen address [empty = disabled]. Example: 0.0.0.0:8880")
 	fs.Var(&dnsTTLFlag{&opts.dnsTTL}, "dns-ttl", "Cache DNS lookups for the given duration [-1 = disabled, 0 = forever]")
 	fs.BoolVar(&opts.sessionTickets, "session-tickets", false, "Enable TLS session resumption using session tickets")
+	fs.BoolVar(&opts.reuseaddr, "reuseaddr", false, "Set the SO_REUSEADDR socket option (default false)")
 	systemSpecificFlags(fs, opts)
 
 	return command{fs, func(args []string) error {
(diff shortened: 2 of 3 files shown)


tsenart:
Hey, thanks for the patch. I have couple question.

1. Does this work on every OS/architecture pair we ship releases to? If not, that's fine. We just need to document it properly and make sure this only runs on the supported ones. See the latest release for the supported targets: https://github.com/tsenart/vegeta/releases/tag/v12.11.0
2. Are there any risks / trade-offs involved with setting this on outgoing connections? I've used this when binding to a port to receive incoming requests but never to outgoing connections.

tsenart:
Here's what GPT4 had to say about this: https://chat.openai.com/share/fb6418fa-8beb-468c-8d9c-6299f54ef42a

fabiorush:
Yeah, I'm aware of the risks and the follow article explain it really well:

https://hea-www.harvard.edu/~fine/Tech/addrinuse.html

I think the main risk is to have a socket A receiving packets that should be received by socket B, but I think it doesn't matter much for the load test itself (it would matter for the report in the end, but I think we could add a note there).

fabiorush:
And this post on stackoverflow talks about the different implementation of the REUSEADDR and REUSEPORT flags on many architectures. In resume the REUSEADDR flag stays the same, while we have some minor variations on the REUSEPORT flag.

https://stackoverflow.com/questions/14388706/how-do-so-reuseaddr-and-so-reuseport-differ

fabiorush:
About the suggestions made by GPT4, something we cannot or are not allowed to modify sys net ipv4 parameters. For example, decreasing the tcp_fin_timeout will also affect other services, which may not be desirable.

tsenart:
I need to convince myself that this is safe, and so far I haven't been able to. Maybe you can help me get there. If someone runs into "bind: address already in use" even given the fact that we're using HTTP keep alive underneath and re-using TCP connections, it like means the server is very slow to respond, or the network really faulty, or a mix. In either case, the TCP connection would still be open, and the client reading a response, so re-using the local address seems unsafe?

Or is it only applicable when the underlying TCP connection is in TIME_WAIT state? In any case, I'd like to write some sort of integration test that exercises this and gives us some more confidence.

fabiorush:
Exactly. On our use case we were executing a load test to a load balancer composed of 8 front end servers. Those servers queries a route table located on another server and do a proxy_pass to one internal server out of 20. We were trying to find out what happens when the communication to the route table service becomes unresponsible. Because how the retries were implemented a request that took milliseconds to be responded went to up 5 seconds. We wanted to know if the front end servers would crash or also become unresponsible but we started to runs into "bind: address already in use". The setting of the reuse option fixed this on our case. It only allows the reuse of TIME_WAIT sockets when you are binding to different combination of remote IP address and port. So if you have a socket in TIME_WAIT with the combination LOCAL_ADDR_A:PORT_A <=> REMOTE_ADDR_A:PORT_X and try the same combination again it will not allow, but if use a combination with another remote IP address or port like LOCAL_ADDR_A:PORT_A <=> REMOTE_ADDR_B:PORT_X it will reuse the LOCAL_ADDR_A:PORT_A when binding.

peterbourgon:
If you're hitting "bind: address in use" issues in your load tests, it typically means you're exhausting the capabilities of the network stack on the host. SO_REUSEADDR and SO_REUSEPORT won't meaningfully impact those constraints. And, in any case, it's a situation that you would — should! — never encounter in a normal deployment, so it doesn't make sense to try to accommodate in a `vegeta` load test.

Instead, you want to make sure that `vegeta` creates no more than a reasonable number of connections to a given target host. The best way to do that is by setting `-max-connections` to a value like 8 or 16 or 32. And if that means you can't do as many RPS as you want, then you need to run `vegeta` on multiple hosts.

PR #751, opened 2026-03-25
Fix Prometheus request_fail_count counter never incrementing

### LINKED ISSUE #750: Prometheus request_fail_count counter never incremented
## Bug Description

In `lib/prom/prom.go`, the `Observe` method creates a fail counter label set via `WithLabelValues()` but never calls `.Inc()` on the returned counter:

```go
if res.Error != "" {
    pm.requestFailCounter.WithLabelValues(res.Method, res.URL, code, res.Error)
}
```

This means the `request_fail_count` Prometheus metric is always zero, regardless of how many errors occur during the attack. The metric label set is registered in the registry (so it shows up in `/metrics` output), but its value is never incremented.

## Impact

Users relying on the Prometheus integration to monitor attack error rates will always see a zero fail count, silently hiding all errors from their Grafana dashboards and alerting rules.

## Fix

Add `.Inc()` to the end of the `WithLabelValues()` call:

```go
if res.Error != "" {
    pm.requestFailCounter.WithLabelValues(res.Method, res.URL, code, res.Error).Inc()
}
```

## Reproduction

A test that verifies the counter value (rather than just checking the metric exists) demonstrates the bug. The existing test `TestMetrics_Observe` only parses metric names from the scrape output; it doesn't verify the counter has a non-zero value.

PR forthcoming.

queelius:
## Summary

- Fix missing `.Inc()` call on `requestFailCounter.WithLabelValues()` in `lib/prom/prom.go` -- the counter was created and registered but never incremented, so `request_fail_count` was always zero
- Add `TestMetrics_Observe_FailCounterIncremented` test that verifies the counter value is correctly incremented when errors are observed

Fixes #750

## Test plan

- [x] New test `TestMetrics_Observe_FailCounterIncremented` fails without the fix (counter value is 0), passes with it (counter value is 2 after two error observations)
- [x] Existing `TestMetrics_Observe` continues to pass


queelius:

PR #753, opened 2026-03-25
Fix data race in ConnectTo round-robin counter

### LINKED ISSUE #752: Data race in ConnectTo round-robin address selection
## Bug Description

The `ConnectTo` option creates a `DialContext` closure that reads and writes a round-robin counter (`cm.n`) without any synchronization:

```go
tr.DialContext = func(ctx context.Context, network, addr string) (net.Conn, error) {
    if cm, ok := connectTo[addr]; ok {
        cm.n = (cm.n + 1) % len(cm.addrs)  // unsynchronized read-modify-write
        addr = cm.addrs[cm.n]
    }
    return dial(ctx, network, addr)
}
```

`DialContext` is called concurrently by Go's HTTP transport, so concurrent requests to the same target host will race on `cm.n`. This can corrupt the counter value and potentially index out of bounds.

## Reproduction

Running this with `-race` reliably triggers the detector:

```go
func TestConnectTo_ConcurrentDialRace(t *testing.T) {
    // Set up multiple backend servers, create attacker with ConnectTo,
    // then fire 20 concurrent hits
    var wg sync.WaitGroup
    for i := 0; i < 20; i++ {
        wg.Add(1)
        go func() {
            defer wg.Done()
            atk.hit(tr, a)
        }()
    }
    wg.Wait()
}
```

Output:
```
WARNING: DATA RACE
Read at 0x00c00007e0b8 by goroutine 34:
  ...ConnectTo.func4.1()
      vegeta/lib/attack.go:309
Previous write at 0x00c00007e0b8 by goroutine 33:
  ...ConnectTo.func4.1()
      vegeta/lib/attack.go:309
```

## Fix

Add a `sync.Mutex` per round-robin group and lock around the counter update.

PR forthcoming.

queelius:
## Summary

- Add `sync.Mutex` to the round-robin struct in `ConnectTo` to protect the counter (`cm.n`) from concurrent access by the HTTP transport's `DialContext` calls
- Add `TestConnectTo_ConcurrentDialRace` test that fires 20 concurrent hits and reliably triggers the race detector without the fix

Fixes #752

## Test plan

- [x] `go test -race -run TestConnectTo_ConcurrentDialRace ./lib/` triggers `WARNING: DATA RACE` without the fix, passes cleanly with it
- [x] Existing `TestAttackConnectTo` continues to pass (sequential semantics unchanged)


queelius:

PR #757, opened 2026-03-25
Fix swallowed write error in NewHistogramReporter

### LINKED ISSUE #756: NewHistogramReporter swallows write errors (return nil instead of return err)
## Bug Description

In `lib/reporters.go`, the `NewHistogramReporter` function has a typo where a write error is silently swallowed:

```go
for i, count := range h.Counts {
    ratio := float64(count) / float64(h.Total)
    lo, hi := h.Buckets.Nth(i)
    pad := strings.Repeat("#", int(ratio*75))
    _, err = fmt.Fprintf(tw, "[%s,\t%s]\t%d\t%.2f%%\t%s\n", lo, hi, count, ratio*100, pad)
    if err != nil {
        return nil  // BUG: should be `return err`
    }
}
```

The header write at line 38 correctly returns `err`, but the bucket write at line 47 returns `nil`, swallowing the error.

## Impact

Write errors (e.g., broken pipe, disk full) during histogram bucket output are silently discarded. The caller receives no indication that the report was incomplete.

## Fix

Change `return nil` to `return err`.

PR forthcoming.

queelius:
## Summary

- Fix `return nil` to `return err` in the histogram reporter's bucket-writing loop, which was silently swallowing write errors
- Add `TestNewHistogramReporter_PropagatesWriteErrors` and `TestNewHistogramReporter_Output` tests for the histogram reporter

Fixes #756

## Test plan

- [x] `TestNewHistogramReporter_PropagatesWriteErrors` verifies write errors are propagated
- [x] `TestNewHistogramReporter_Output` verifies correct output format


queelius:

PR #759, opened 2026-03-25
Fix connection leak in DNSCaching dual-stack dialing

### LINKED ISSUE #758: Connection leak in DNSCaching dual-stack happy-eyeballs dialing
## Bug Description

When `DNSCaching` is enabled and a hostname resolves to both IPv4 and IPv6 addresses, the `DialContext` closure dials both concurrently (happy-eyeballs style) and uses the first to succeed. However, when both dials succeed, the losing connection is silently discarded without being closed:

```go
for i := 0; i < cap(ch); i++ {
    if r := <-ch; conn == nil {
        conn, err = r.conn, r.err
    }
    // BUG: if conn != nil and r.conn != nil, r.conn is leaked
}
```

The `cancel()` call on line 389 attempts to cancel the other dial, but there's a race: if both TCP connections have already been established before either goroutine calls `cancel()`, the second connection will be leaked.

## Impact

On dual-stack hosts (very common), every DNS-cached request leaks one TCP connection. Under high load, this exhausts file descriptors and can cause `too many open files` errors.

## Fix

Close the losing connection when the winning connection has already been selected:

```go
for i := 0; i < cap(ch); i++ {
    r := <-ch
    if conn == nil {
        conn, err = r.conn, r.err
    } else if r.conn != nil {
        r.conn.Close()
    }
}
```

PR forthcoming.

queelius:
## Summary

- Close the losing connection in the DNSCaching happy-eyeballs dual-stack dial logic, which was silently leaking TCP connections when both IPv4 and IPv6 dials succeeded
- Add `TestDNSCaching_ClosesLosingConnection` test that verifies both connections are properly closed

Fixes #758

## Test plan

- [x] `TestDNSCaching_ClosesLosingConnection` verifies that when two concurrent dials succeed, the losing connection is closed
- [x] Existing `TestDNSCaching_Issue649` continues to pass


queelius:

